PR #4, opened 2017-06-13
Feature/CTRL_L

aaxu:
I implemented this feature to select all the current line that the cursor is on. Subsequent uses of the shortcut will also select subsequent lines (rather than de-selecting a selected line). 

I'm not sure what branch you prefer for me to merge my pull requests into, so I'm guessing it'll be to the master branch rather than releases. Let me know if you need anything changed or if there's anything I should do in future contributions. Thanks!


aaxu:
I just read your comment to #1 and I think that for this function **selectCurrentLine**, I should be using cursorRow to determine what row the cursor is in. However, I'm not sure how to gain access to that variable. I might need to pass the window itself to its controllers and create a parent field so its fields can be seen. Maybe we'll have to get to refactoring soon, since it seems that I'm doing a bit of a hack as well.

I would also have liked to put the cursor at the beginning of the next available line so that it makes more sense to the user, but it doesn't seem that that is possible without a workaround do to how the code is structured (ie: can't move cursor down without being affected by the selection mode)

app/cu_editor.py
```diff
@@ -32,6 +32,8 @@ def initCommandSet(editText, textBuffer):
 
     CTRL_H: textBuffer.backspace,
 
+    CTRL_L: textBuffer.selectCurrentLine,
```
dschuyler: Is the intent to have ctrl+L operate everywhere, including the prompt UIs at the bottom of the screen (like Find and Goto line)? If so this is the spot for this line.

If not (like if the intent is for ctrl+L line selection to be in the main edit window) then please move this down to the CuaPlusEdit class around line 312.

app/help.py
```diff
@@ -96,5 +96,6 @@
   See readme.md for an introduction
   Both files may be found in "%s"
   Please send feedback and bug reports to dschuyler@
+  (Release build)
```
dschuyler: Oh, this is an intentional difference between the master and release branches. It's purpose is to tell us when a user is using a stable release build vs a non-stable master branch build. Please don't add this to the master branch.

app/text_buffer.py
```diff
@@ -1289,6 +1289,17 @@ def doSelectionMode(self, mode):
           mode - self.selectionMode)))
       self.redo()
 
+  def selectCurrentLine(self):
+    """
+      This function is used by the CTRL + L shortcut.
```
dschuyler: Let's keep these text_buffer.py function comments more generic. Please remove this line about CTRL + L. I.e. keep the bit about what the function does, without referring to how it is triggered. 

The reason is that vi-like bindings or emacs-like bindings may use different bindings.

app/text_buffer.py
```diff
@@ -1289,6 +1289,17 @@ def doSelectionMode(self, mode):
           mode - self.selectionMode)))
       self.redo()
 
+  def selectCurrentLine(self):
```
dschuyler: Consider naming this cursorSelectLine. This will be consistent with a collection of cursorSelect* functions.

(The naming will make more sense once cursorRow/Col is correctly separated from penRow/Col).

dschuyler:
The master branch is the right branch to path too, thanks!

app/cu_editor.py
aaxu: Yes the intent is for it to operate everywhere, since there doesn't yet seem to be an easy way to erase the text written in the prompt UIs at the bottom of the screen.

app/help.py
aaxu: Oh sorry, I think I made it branch from the release branch because I didn't know which of the branches I was going to merge into. I'll start making everything based on the master from now on! 👍

aaxu:
Changes have been resolved!

dschuyler:
Awesome, thanks!

dschuyler:
> I should be using cursorRow to determine what row the cursor is in

True. Your assessment about that not being practical is also true. The separation of cursor and pen is still in progress. Using pen is good for now (since opening multiple views on the same buffer is not a current feature - but it's a planned future feature).

> I would also have liked to put the cursor at the beginning of the next available line

That gets tricky at the last line of the doc. There would need to be a special case for the last line or the last line would not be selectable (which felt like a bug when I did it :) ). 

Long term, I'm considering removing the cursor when a selection is made. The idea being that the whole selection is equivalent to the cursor. Showing the cursor in addition to the selection is somewhat redundant. I'm up for discussing it, this is just the direction I'm currently considering.

aaxu:
> Long term, I'm considering removing the cursor when a selection is made. The idea being that the whole selection is equivalent to the cursor. Showing the cursor in addition to the selection is somewhat redundant.

This is an interesting idea, which I never thought of. The only con about it is that when you have a selection and you want to exit out using the arrow keys, you might not know where you end up? Although I guess we can just follow some standards. For example, given that the user has a selection:
* If the left or right arrow key is pressed, then move the cursor to the respective end of the selection.
* If the up or down arrow key is pressed, move the cursor (somewhere) in the line above or below the selection, if a line exists.

I think it's a neat and clean idea :+1:

PR #8, opened 2017-06-15
Bug fix/universal key codes

aaxu:
I made most of the constants in commandSets as strings because I think the integer representation has multiple types of values that it can take (depending on the mode of the terminal and other various settings). Some keycodes could not be turned into keynames (for example, the ALT + A, ALT + B, ALT + C, and ALT + S keys), so I left them as integers. The doCommand function now checks whether the string representation or the integer representation of an input is inside the commandSet since the dictionary now may contain a mix of both.

One thing is that I could not test this on Mac so I left the keycodes in the 'darwin' section in curses_util.py as is until @dschuyler or someone else can check whether the new keycodes work on macOS as well.


app/controller.py
```diff
@@ -66,8 +66,10 @@ def changeToSaveAs(self):
     self.host.changeFocusTo(self.host.interactiveSaveAs)
 
   def doCommand(self, ch):
+    #Check the commandSet for the input with both its string and integer representation.
     self.savedCh = ch
-    cmd = self.commandSet.get(ch)
+    keyName = app.curses_util.cursesKeyName(ch)
```
dschuyler: Thanks for working on this. I'm not following a part of this though: what would happen if strings were not used and integers (and tuples) were used? What bug or issue would we see?

aaxu: I'm not sure what you mean by that, so tell me if this answers your question. 

* In the case that an integer is used, the corresponding key name (string) would be found and we can access the command that it corresponds to in the commandSet dictionary. 

* If a tuple is used, as it sometimes is (I've only encountered this for some ** ALT + CHAR** keys so far in some OS or terminals), nothing would happen because **cmd** in **doCommand()** would be **None** (as the tuple currently isn't in the dictionary), and so it will default to **defaultCommand()**. The default command for most of the controllers is **insertPrintable**. This function calls the function **curses.ascii.isprint(ch)** which returns whether a character is printable or not. In this case, **ALT + CHAR** is not printable so the function would exit and nothing would be printed. This behavior is also the same for if some integer keycode and it's string representation (if it has one) is not in the dictionary. For these **ALT + CHAR** shortcuts, I'm not sure how to go about them yet since they seem to be correct as integers for your system, but are tuples on my system. I'm currently working on figuring out a way to make the conversion work, so hopefully that goes well.

In summary, there should be no other bugs, other than the fact that some commands would not work if the keycodes or keynames are not the same for some particular system. I'd be good if you could confirm on your systems whether or not all the commands work

aaxu: Actually, after some more thought, this is still hard coding in the values for these keys, so I don't think it will work across all platforms and terminals. I may have to try and see if it's possible to simulate these keys being pressed at the beginning of a session and store these values for future use (maybe save it as a file as well?). This should solve the problem of differing key values if python has a way to allow such a method.

aaxu:
Hey @dschuyler, I've been stuck on this problem for a little bit and need a little bit of help. I've discovered that the curses.keyname() function in the curses library actually has some other way to determine the keyname of a given integer. For example, on one system, 543 is given the keyname of 'kLFT3' on system1, but 546 is 'kLFT3' on system2. Thus, I wanted to stick by the curses defined constants/keynames of possible which is why most of the values are strings except some defined curses constants like curses.KEY_HOME.

The thing that I noticed is that backspace has only 3 keys that it can vary across terminals and platforms. They are 127, 263, and 8 (which is usually used for CTRL + H). I added all these as possibilities for backspace. However, CTRL + Backspace may also occupy one of these values, so I'm not really sure how to differentiate them at the moment. That was going to be my next implementation, but I don't know if it's possible right now :(.

Other than that, the thing that I need help with is the ALT + A, B, C, S keys. In both of my systems, they return a tuple **(ch, )** where ch is the ascii code of the character pressed with alt. I'm not sure if this is the case for you so I was hoping you could check that it's the same for your systems. I tried looking at the source code of getch, but I couldn't figure out how to get it to detect ALT as it's own key or why it is weirdly returning a tuple. Let me know if the same thing is happening on your linux system.

dschuyler:
Thanks again for looking into this. 

I made [better comments in the commandLoop](https://github.com/google/ci_edit/blob/master/app/ci_program.py#L109) function that may explain the "returning a tuple". If that's not explaining it, please let me know (so I can fix the explanation).

I might not be possible to get the ALT key on it's own unless the terminal will send that information. Many older terminals refuse to send modifier keys unless they are accompanied by a non-modifier key (or said the other way around, they send key events and those key events may have modifiers).

I'd really like a terminal that sends key down and key up events for each button on the keyboard, but I haven't found that terminal yet.

I'll follow-up with more info on key code results that I see.

aaxu:
The weird thing is that I was researching this and apparently calling curses.raw() is supposed to make it so that getch() is able to detect the ALT key with an integer code of 27 which  we can then use as kind of a key down function for the modifier keys. However, it seems that we already have curses.raw() called, but it's not detecting the modifier keys by themselves.

Also, there is an curses.ascii.alt(c) function which returns the ascii value of the alt modified version of a pressed character. If you know how to convert the tuples back into an integer and if they are equal to each other, then it may work out!

dschuyler:
FYI: Here are some mac os key codes using iterm2 in the Key Mappings I normally use:

alt+a = 165 M-%
alt+b = 171 M-+
alt+c = 167 M-'
alt+s = 159 M-^_

alt+<left arrow> = esc + (91, 49, 59, 57, 68)
alt+<right arrow> = esc + (91, 49, 59, 57, 67)

Pressing alt on its own (or any modifier key on its own) does not generate a getch() result.

iterm2 will allow loading various key settings, so for example in 
- term.app 
    - shift+up = (91, 65)
    - ctrl+up = (91, 65)
    - which means we can't tell them apart.
- xterm.defaults
    - shift+up = 337 KEY_SR
    - ctrl+up = (91, 49, 59, 57, 65)
    - so we can tell them apart

dschuyler:
If I understand correctly, we are at the mercy of the terminal (or terminfo?) to decide what to send us.

app/ci_program.py
```diff
@@ -105,7 +105,7 @@ def commandLoop(self):
       while not len(cmdList):
         for i in range(5):
           ch = window.cursorWindow.getch()
-          if ch == curses.ascii.ESC:
+          if ch == app.curses_util.KEY_ESCAPE:
```
dschuyler: I think this one should stay as curses.ascii.ESC because it's testing for the start of an escape sequence which has a particular value.

app/ci_program.py
```diff
@@ -121,7 +121,7 @@ def commandLoop(self):
               # The sequence was empty, so it looks like this Escape wasn't
               # really the start of a sequence and is instead a stand-alone
               # Escape. Just forward the esc.
-              ch = curses.ascii.ESC
+              ch = app.curses_util.KEY_ESCAPE
```
dschuyler: But this one can be app.curses_util.KEY_ESCAPE because we're no longer looking for an escape sequence.

app/controller.py
```diff
@@ -66,8 +67,10 @@ def changeToSaveAs(self):
...
+    cmd = self.commandSet.get(keyName) or self.commandSet.get(ch)
```
dschuyler: How about changing this to look for the integer .get(ch) first and then if that fails, do the cursesKeyName(ch) and the .get(keyName). The reason is that it's less work for the cpu.

app/cu_editor.py
```diff
@@ -75,8 +74,8 @@ def __init__(self, host, textBuffer):
     self.document = host
     commandSet = initCommandSet(self, textBuffer)
     commandSet.update({
-      ord('n'): self.closeFile,
-      ord('y'): self.writeOrConfirmOverwrite,
+      'n': self.closeFile,
+      'y': self.writeOrConfirmOverwrite,
```
dschuyler: I think these two lines can/should stay as ord() since they are looking for the ascii value.

app/cu_editor.py
```diff
@@ -90,7 +89,7 @@ def __init__(self, host, textBuffer):
     self.document = host
     commandSet = initCommandSet(self, textBuffer)
     commandSet.update({
-      ord('y'): self.overwriteHostFile,
+      'y': self.overwriteHostFile,
```
dschuyler: Same, can we go with ord()

app/curses_util.py
```diff
@@ -75,36 +91,52 @@
...
-  KEY_CTRL_SHIFT_RIGHT = 560
-  KEY_CTRL_UP = 565
-  KEY_CTRL_SHIFT_UP = 566
+  KEY_ALT_LEFT = 'kLFT3'
+  KEY_ALT_RIGHT = 'kRIT3'
+  KEY_ALT_SHIFT_LEFT = 'kLFT4'
+  KEY_ALT_SHIFT_RIGHT = 'kRIT4'
+  KEY_CTRL_DOWN = 'kDN5'
+  KEY_CTRL_SHIFT_DOWN = 'kDN6'
+  KEY_CTRL_LEFT = 'kLFT5'
```
dschuyler: Keen, on darwin ctrl+left sends 541 and on Linux it's 544, but both convert to kLFT5 -- which may be exactly what that string label is trying to help with (make it so we don't have to workout all the codes). Guessing, but seems that seems reasonable.

app/ci_program.py
aaxu: Hmm yeah you're right. I was thinking that it would be okay since I assigned the same value to the new variable, but this would break if we end up changing the value later on.

app/ci_program.py
aaxu: This should be the same as above right? Since we know for sure we're going to be searching for an integer value.

app/controller.py
aaxu: Oh yeah I was thinking about this, but I decided to do it the other way because a lot of the commands were strings before. Do you think that I should change back the CTRL + CHAR commands back to an integer value? The reason I ask is because I'm not sure if the ascii value will stay consistent across platforms but the string value seems consistent, at least for control keys.

app/ci_program.py
dschuyler: Oh, yeah. I guess it would work either way here; but if we optimize for integers then an int would be better for that reason. cool.

app/controller.py
dschuyler: I think using int and str together should work well. 

Though I just realized that the optimization I was hoping for will not pan out. I was thinking that a common ch value will be non-modified letters, like the letters I'm typing now in this sentence. Therefore we should check ints first, but I've coded it so that letters are entered as a fall through if no command is found. So, every normal letter typed is going to do a cursesKeyName() lookup, plus two .get() calls; before falling through to a normal letter input.

The calls to .get() are likely super cheap cpu-wise (based on prior tests), but a call into curses has been been slow in the past, so that's what I was trying to avoid. I may be 'optimizing too early'. Maybe I/we should try it and see if its a performance issue or not.

aaxu: Ah you're right, I forgot about printable keys. I'll switch it.

app/ci_program.py
aaxu: I've switched it back to curses.ascii.ESC for now. It maybe slower to access the library, instead of our stored one, but maybe that's just early optimization.

aaxu:
Changes are all done. Maybe I should have edited it via the website to prevent all the outdated reviews.

PR #17, opened 2017-06-29
Pasting Error Fix

aaxu:
An error would print by a subprocess function when you try to paste a non-printable object from the clipboard like an image. This redirects the error message to null. I didn't want to modify this third party library directly, but it didn't seem like it would let me redirect this output in app.clipboard.paste() or text_buffer.editPaste()

### DIFF
--- a/third_party/pyperclip/clipboards.py
+++ b/third_party/pyperclip/clipboards.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import subprocess
 from .exceptions import PyperclipException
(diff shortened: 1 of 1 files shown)


aaxu:
What would the PR be about? It's just displaying an error message, and it lets us redirect the output if we want. Maybe I just don't understand how python works enough to understand why I can't call a function and redirect all output that results from that function call. I guess if anything, it would be some function that automatically suppresses all output in the library when called.

PR #21, opened 2017-06-29
Bug fix for saving in InteractiveFind and InteractiveGoTo

### LINKED ISSUE #20: Saving in InteractiveFind or InteractiveGoTo mode sometimes does not work
This happens when you start up the text editor without specifying a file name. When you enter InteractiveFind or InteractiveGoTo mode and try to save, it will say there was an Error saving the file.

aaxu:
Resolves #20

### DIFF
--- a/app/controller.py
+++ b/app/controller.py
@@ -130,12 +130,14 @@ class Controller:
       self.changeToConfirmOverwrite()
       return
     tb.fileWrite()
+    saveMessage = tb.message #Store the save message so it is not overwritten
     if self.host.userIntent == 'quit':
       self.quitOrSwitchToConfirmQuit()
       return
     if self.host.userIntent == 'close':
       self.closeHostFile()
     self.changeToHostWindow()
+    tb.message = saveMessage #Restore the save message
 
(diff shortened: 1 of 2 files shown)


aaxu:
Found a thing that might be an issue. If you try to do the same steps, but instead a file name was specified, the file still saves, but there is no message that says "file was saved" which causes confusion. Looking into this

app/controller.py
```diff
@@ -130,12 +130,14 @@ def writeOrConfirmOverwrite(self):
       self.changeToConfirmOverwrite()
       return
     tb.fileWrite()
+    saveMessage = tb.message #Store the save message so it is not overwritten
```
dschuyler: Please help me reproduce the situation that saveMessage fixing. I haven't hit it yet. I'm wondering if there's a deeper problem I should address, but I'd like to repro the issue to understand better.

aaxu: Enter the InteractiveFind prompt with CTRL + F. Then type some string that is not in the document so that the message "Find wrapped around" is shown up. Then try to save and it will simply exit interactive find and all messages disappear, so it is unclear whether save worked or not (it did)

Issue #31, opened 2017-07-11
Bug with Finding when there is only one line of text


aaxu:
If you have only one line of text and your cursor is at the end of the line, find will not work properly. Also, regardless of where your cursor is, if you keep pressing findNext, it will have weird behaviors like nothing will be highlighted.

PR #32, opened 2017-07-11
Fixed find bug that occurred when there was only one line of text

aaxu:
I made some changes to how **find** behaves.

* When you press find next or find previous (CTRL_F and CTRL_R respectively), and there is only one such match when there is only one line of text, that one selection will stay highlighted.


aaxu:
Should resolve #31

dschuyler:
Sorry I've been slow to respond this week. I'll try to look at PRs on Friday.

app/text_buffer.py
```diff
@@ -687,6 +687,7 @@ def applyDocumentUpdate(self, data):
 
   def findCurrentPattern(self, direction):
     localRe = self.findRe
+    offset = self.penCol + direction
```
dschuyler: I've moved this function into ci_edit/app/actions.py. Are you up for moving the PR there?

app/text_buffer.py
```diff
@@ -732,15 +732,31 @@ def findCurrentPattern(self, direction):
       theRange = range(self.penRow)
     else:
       theRange = range(len(self.lines) - 1, self.penRow, -1)
-    for i in theRange:
-      found = localRe.search(self.lines[i])
+    if theRange:
```
dschuyler: Not a big deal, but I wondered if lines 753 to 759 can be combined with 736 to 744 since they are similar. Maybe line 735 could be 

if not theRange:

and then theRange could be made to work with the 'for i in theRange'? I haven't carefully reviewed whether that would yield tighter code - so if it doesn't, then ignore this review comment. My gut is that it's likely to work out though.

Either way, thanks for address this. I'm looking forward to merging it.

app/text_buffer.py
aaxu: Of course!

app/text_buffer.py
aaxu: I cleaned up the code a little bit. It now repeatedly checks if found is ever not None and if not, it will keep trying.

aaxu:
Requested changes have been completed.

PR #33, opened 2017-07-12
Made find case insensitive by default.

aaxu:
I added a new instance variable to BackingTextBuffer called self.ignoreCaseFind. It is set to true by default and makes finding case insensitive. I was thinking that when we end up implementing a button that turns this off/on, we could simply switch the value of this variable and call find() again. I would do this, but I'm not really sure how to implement a button in curses with some kind of pressed down animation.


dschuyler:
Sorry I've been slow to respond this week. I'll try to look at PRs on Friday.

app/text_buffer.py
```diff
@@ -607,8 +610,12 @@ def find(self, searchFor, direction=0):
       self.doSelectionMode(app.selectable.kSelectionNone)
       return
     # The saved re is also used for highlighting.
-    self.findRe = re.compile('()'+searchFor)
-    self.findBackRe = re.compile('(.*)'+searchFor)
+    if self.ignoreCaseFind:
+      self.findRe = re.compile('()'+searchFor, re.IGNORECASE)
```
dschuyler: wdyt of a flags variable set to something like

flags = self.ignoreCaseFind and re.IGNORECASE or 0
self.findRe = re.compile('()'+searchFor, flags)
self.findBackRe = re.compile('(.*)'+searchFor, flags)

and not having the if/else?

app/text_buffer.py
```diff
@@ -40,6 +40,9 @@ def __init__(self):
     self.rootGrammar = app.prefs.getGrammar(None)
     self.skipUpdateScroll = False
 
+    # Determines whether case matters when finding a selection
+    self.ignoreCaseFind = True
```
dschuyler: how about adding 

'findIgnoreCase': True,

in ci_edit/app/default_prefs.py in the 'editor' section? It could be checked here with

app.prefs.prefs['editor'].get('findIgnoreCase')

That will allow us/users to set

"findIgnoreCase": false,

in ~/.ci_edit/prefs/editor.json to change it.

app/text_buffer.py
```diff
@@ -607,8 +610,12 @@ def find(self, searchFor, direction=0):
       self.doSelectionMode(app.selectable.kSelectionNone)
       return
     # The saved re is also used for highlighting.
-    self.findRe = re.compile('()'+searchFor)
-    self.findBackRe = re.compile('(.*)'+searchFor)
+    if self.ignoreCaseFind:
```
dschuyler: I've moved the BackingTextBuffer into ci_edit/app/actions.py, are ya cool with making the changes there?

app/text_buffer.py
aaxu: Yeah that would make it a lot cleaner. I'll make these changes

app/text_buffer.py
aaxu: Oh yeah for sure, I'll update this PR

Issue #27, opened 2017-07-05
Tab usage


aaxu:
I was thinking about how the tab key works and wanted your opinion on some things @dschuyler.

* When the selection mode is none, should we just insert a tab and move the cursor to the end of it rather than indent the whole line and have the cursor at the beginning of the line?

* When the selection mode is not none and we tab, should we keep the cursor at the current position or keep it moving to the beginning of the line?

* Let's say that there is a colon and then a new line. The new line operator will put the cursor on an indented spot on the next line, a few spaces after the beginning of the previous line. Should we have tab jump to this position if the cursor is before this position?
For example:

`for x in lst:`
`xx|yyyyyy`

If the cursor is in some position with an x, them jump to the vertical bar if a tab is inserted (all selection modes).

dschuyler:
Good ideas. (though I'm not sure I follow the third bullet point).

- If we add tab insertion, we should also add the option to insert tabs or spaces. E.g. the coding standard I use at work requires two space indentation (and no tabs) and it's really troublesome if the editor inserts tabs. Though some other employer may require tabs and it would be great for ci_edit to work for those users as well.

On the third bullet point: 

- Would pressing tab a second time then insert a tab (i.e. if the cursor is already at the first non-whitespace character or eol)?

- I've noticed that some editors will do something similar iiuc, with the HOME key. e.g. there are two 'home' positions. One is at the first character of the line (textBuffer.line[N][0]) and the other is at the first non-blank character (or eol). Pressing HOME repeatedly will move the cursor between the two home positions.

aaxu:
Yeah I think the first bullet is a good idea and the third bullet is an interesting one that I haven't heard of. I think that sounds fun to use and would like to see it implemented. Maybe I'll take it on soon if I can't find any other bugs.

Also, for the second bullet point, I think it logically makes sense to insert a tab the second time. I can't think of any other functions that it could have since the first tab should put the line at supposedly the correct indentation level.

aaxu:
What do you think backtab should do when no text is selected? Should it do a simple tab, do nothing, or have some other function? It doesn't seem exactly perfect to make it unindent the entire line when a regular tab simply inserts a tab at the cursor's current position.

PR #34, opened 2017-07-18
Tab usages

aaxu:
I changed the functions of tab so that it now has the following features:
 * When no text is selected, tab will insert an indentation at the cursor's position (default is 2 spaces).
* When text is selected, tabbing will indent the beginning of all lines that the selected text is in, even if a full line is not highlighted. 
* When indenting multiple lines, the highlighted area will stay in the same position relative to the text it is around. This means that whatever is highlighted will stay highlighted and whatever is not highlighted will not be highlighted

* Back tabbing currently has no function if nothing is highlighted (I'm not sure what function it should have so let me know what you think it should be).
* Back tabbing for all other selection modes will work as regular indenting does except you will unindent instead of indenting (provided that there is an indent at the beginning of the selected lines).


app/actions.py
```diff
@@ -775,29 +777,29 @@ def findPrior(self, searchFor):
...
-          self.penRow - self.markerRow, self.penCol - self.markerCol, 0)
-      self.redo()
-      self.indentLines()
-    elif self.selectionMode == app.selectable.kSelectionAll:
-      self.cursorMoveAndMark(len(self.lines) - 1 - self.penRow, -self.penCol,
-          -self.markerRow, -self.markerCol,
-          app.selectable.kSelectionLine - self.selectionMode)
-      self.redo()
-      self.indentLines()
+      self.verticalInsert(self.penRow, self.penRow, self.penCol, indentation)
```
dschuyler: I think this doesn't need to be a verticalInsert, though I suppose it doesn't hurt.

app/actions.py
```diff
@@ -177,6 +177,8 @@ def cursorMove(self, rowDelta, colDelta):
 
   def getCursorMoveAndMark(self, rowDelta, colDelta, markRowDelta,
       markColDelta, selectionModeDelta):
+    if self.penCol + colDelta < 0: #Catch cursor at beginning of line
```
dschuyler: Trivia: I've been trying for a style similar to the Chromium style guide, though I haven't noted that yet. In that style there is two spaces before an end of line comment, a space after the # and a period at the end.

Like:

if self.penCol + colDelta < 0:  # Catch cursor at beginning of line.

app/actions.py
```diff
@@ -1062,28 +1064,26 @@ def stripTrailingWhiteSpace(self):
...
 
   def unindent(self):
-    if self.selectionMode == app.selectable.kSelectionAll:
-      self.cursorMoveAndMark(len(self.lines) - 1 - self.penRow, -self.penCol,
-          -self.markerRow, -self.markerCol,
-          app.selectable.kSelectionLine - self.selectionMode)
-      self.redo()
-      self.unindentLines()
+    if self.selectionMode == app.selectable.kSelectionNone:
+      pass
```
dschuyler: Maybe a comment about whether this is a TODO for later or is an intentional no-op.

dschuyler:
Thanks!

app/actions.py
aaxu: Yeah it seemed a lot cleaner and since I made indent lines only for indenting the beginning of all lines (when you tab with a a non-none selection mode), I just used vertical insert. Both functions use verticalInsert and verticalInsert wasn't really used for anything yet so I figured it was okay.

app/actions.py
aaxu: Oh yeah sorry, I'm still learning all the styles and stuff, picking stuff up here and there. Should we start doing that from now on? The google python style guideline specifies 2 spaces before a comment so it would be 

if self.penCol + colDelta < 0:  # Catch cursor at beginning of line.

Mix and match? :+1:

app/actions.py
aaxu: Oh I asked you about this in #27 so I intentionally left this blank for now. I'll update that my next PR if you want.

PR #36, opened 2017-07-28
Add redo chain to history

aaxu:
**Features**
* Now, user history is saved when you save a file, rather than when you exit the program.
* When you re-open the file, you will be in the exact same state as before. Highlights, cursor, selectionMode, and redo chains are all preserved, provided that the file was not modified by another program.
* The way I match a file to its history is hashing it using a sha512 function as well as check its file size to see if it was the same as before. This reduces the chance of a collision and also allows the matching to occur even if the user moves the file to another location.


app/window.py
```diff
@@ -873,21 +873,10 @@ def refresh(self):
   def setTextBuffer(self, textBuffer):
     app.log.info('setTextBuffer')
     #self.normalize()
-    if self.textBuffer is not None:
-      app.history.set(['files', self.textBuffer.fullPath, 'cursor'],
-          (self.textBuffer.penRow, self.textBuffer.penCol))
-    # TODO(dschuyler): Do we need to restore positions and selections here?
     self.controller.setTextBuffer(textBuffer)
     Window.setTextBuffer(self, textBuffer)
+    self.textBuffer.restoreUserHistory()
```
dschuyler: I wondered why this was here rather than were the file is read (for symmetry with the file save code in this PR. Is it for the self.view.scroll* values? i.e. so that self has a self.view?
Consider not saving/restoring the scroll position. Try instead scrolling to the cursor position on open. If that works well, maybe this can be moved into the file loading code.

app/actions.py
```diff
@@ -551,28 +553,56 @@ def fileLoad(self):
...
+        ['files', self.fullPath, 'scroll'], (self.view.scrollRow, self.view.scrollCol))
+    app.history.set(
+        ['files', self.fullPath, 'marker'], (self.markerRow, self.markerCol))
+    app.history.set(
+        ['files', self.fullPath, 'selectionMode'], self.selectionMode)
     # Preload the message with an error that should be overwritten.
     self.setMessage('Error saving file')
     try:
       try:
-        self.stripTrailingWhiteSpace()
```
dschuyler: Ah, I can see were this may be odd. I rely on this though. WDYT of making this a preference option instead of removing it?

app/default_prefs.py
```diff
@@ -517,4 +518,8 @@ def numberTest(str, expectRegs):
       "backgroundIndexes": [231, 229, 14, 221,   255, 254, 253, 225],
     },
   },
+  'userData': {
+    'homePath': os.path.expanduser('~/.ci_edit'),
+    'historyPath': os.path.join(os.path.expanduser('~/.ci_edit'), 'history.dat'),
+  },
```
dschuyler: Nice!

app/actions.py
dschuyler: Hmm, it also looks like it may create a lot of things to undo :(. Maybe that should be a single undo event rather than a bunch.

app/window.py
aaxu: Yeah it was primarily because there was no view that I moved it here. I'll see if I can scroll to the cursor, but I think its better if we center the cursor maybe?

app/actions.py
aaxu: Yeah the issue I had with this was that it created a lot of extra entries in the redo chain, so when I saved the redoChain, it didnt seem to match. I can fix this though by just saving the redo chain after stripping the white space. That might cause confusion upon reloading the file since there are now commands that the user can undo that they themselves didnt perform. What do you think about making this a shortcut instead of a setting?

app/default_prefs.py
aaxu: Still need to work on bookmarks later on but I'm still not really sure what that does yet LOL

app/actions.py
dschuyler: I made some changes to try to help with this. 
* added Compound Changes so that multiple actions/mutations can be grouped together as a unit
* made onSaveStripTrailingSpaces() use a Compound Change
* added 'onSaveStripTrailingSpaces' editor preference so that users may choose whether that happens when saving

app/window.py
dschuyler: I try to put the current cursor line ten lines down from the top. Maybe we can make that a percentage and a user pref? Like I could set it to 25% because I like it nearer the top and you could set it to 50% to center it. WDYT?

dschuyler: I checked in an example, search for optimalCursorRow to see it.

app/history.py
```diff
@@ -19,44 +19,65 @@
...
 import cPickle as pickle
 import os
+import hashlib
 
-data = {}
-path = None
+fileHistory = {}
+userHistory = {}
+pathToHistory = None
+hasher = None
```
dschuyler: It looks like this |hasher| can be removed.

app/history.py
```diff
@@ -19,44 +19,65 @@
 import app.log
 import cPickle as pickle
 import os
+import hashlib
 
-data = {}
-path = None
+fileHistory = {}
+userHistory = {}
+pathToHistory = None
```
dschuyler: How about if the code that uses |pathToHistory| use app.prefs.prefs['userData']['historyPath'] instead?

app/history.py
```diff
@@ -19,44 +19,65 @@
...
+fileSize = 0
 
-def get(keyPath, default=None):
-  global data
-  cursor = data
-  for i in keyPath[:-1]:
-    cursor = cursor.setdefault(i, {})
-  return cursor.get(keyPath[-1], default)
+def get(keyPath, default={}):
+  return fileHistory.setdefault(keyPath[-1], default)
```
dschuyler: This looks like it's only using the last element of keyPath. What's the advantage (or is this a whoops)?

app/window.py
aaxu: Hmm I guess it's good to make it preferable. Although I would assume that the user is working at a comfortable scroll view so when they resume editing a file, the position should be saved. What about making it a setting with default as saved?

app/history.py
aaxu: Yeah I don't really know. When I first started this PR, history and bookmarks were saved with keys that were like ['files', self.fullPath, **attribute here**]. I didn't know if you had a reason, but I just kept it for now. I can drop all these and just have a string as the key instead of an array if the rest doesn't have any purpose.

app/window.py
aaxu: I think that this idea is really good for when you snap back to the cursor when your scrollView is at a different location. This currently is only implemented for when you find an expression and the result is not in the view. What do you think about making this optimalCursorRow/Col to apply also when your cursor is out of view and you type/use the arrow keys? 

I still think when specifically resuming a file, the position should be saved at the same spot? Or maybe we can cap it at the optimalCursorPos defaults. For example, if the cursor is at below the preference (.28) and I save and resume editing, the view will now be scrolled to .28, but if the cursor was at .48 (percentage of rows from the bottom), then it would stay at .48

app/history.py
aaxu: Sounds good!

app/history.py
aaxu: Ah you're right. Changed!

dschuyler:
Thanks for making updates to the PR!

On the scroll position, my concern has to do with opening the file with a different 'shaped' window. e.g. imagine that the window is 80 lines and the scroll position is 100: so we're seeing lines 100 to 179. The pen row is 170. If we reopen with the same window shape, all is well. but imaging we open in a window that is 30 lines tall: so we're seeing lines 100 to 129. The pen is still down on line 170. Even though a portion of the screen that was on the screen when it was saved is still on the screen, it's likely that out eye focus was around line 170. Now the middle of the window has our eye around line 115 (middle of the screen) and we're wondering, "what are we looking at?".

So it may work to save the scroll position and window shape (rows, cols) and if the window shape is the same, restore the scroll position. If the window shape is different, toss out the scroll position and make an educated guess at what the user would find helpful to frame on the screen (i.e. the pen position).

That's more work to do, but it would address the concerns.

I was suggesting skipping the first part and making an educated guess straight out of the gate.

I may also suggest doing the scroll to cursor position for the first-pass PR and do the fancier scroll restoration in a follow-up PR (totally up to you, but I'm a fan if incremental, smaller changes).

WDYT? (I'm open to other concerns/suggestions -- as always).

aaxu:
Oh I see that does pose a problem. Yeah okay I'll get the basic framework to just finish up this PR first then.

aaxu:
Sorry for the delay. I made a quick implementation for the optimal cursor position. I plan to make another PR for a smarter way to restore the cursor position after this.

Also, is there some variable that keeps track of the number of lines that are displayed at a time on the screen? Right now,  I'm using self.view.rows which gives a value that is 2 more than the number of lines displayed. However, I'm not sure if this is bound to change in the future so I wanted a variable which I know for sure keeps track of this value.

dschuyler:
It should be self.view.rows. I think that had a bug that was fixed recently. Is it working now?

aaxu:
Hmm that's actually what I thought. However, when the program loads up and **restoreUserHistory** is called. self.view.rows is actually 2 more than what it's supposed to. Weird..

dschuyler:
Ah, it may be a timing thing. The text area starts out taller and is reduced for the TopInfo lines (the file path and the context line). In theory the intention is that the height of the TopInfo may be dynamic. Though so far it's set to two and stays at two.

aaxu:
Hmm then what about setting the initial value of self.topRows to 2? That should fix the timing of it. It appears as though the initial value doesn't matter as long as it's positive.

aaxu:
It's not a huge deal, but it just makes things look cleaner. Like if someone were to look at the - 3 in the code, it's pretty hard to understand that without a deeper understanding of the program.

app/history.py
```diff
@@ -19,44 +19,65 @@
...
-    path = filePath
-    if os.path.isfile(path):
-      with open(path, "rb") as file:
-        data = pickle.load(file)
-    else:
-      data = {}
-  except Exception, e:
-    app.log.error('exception')
-    data = {}
+fileHistory = {}
```
dschuyler: It looks like fileHistory expects that a single file will edited at a time. This is currently true, but I'd like to expand ci_edit to allow multiple files being edited. How about if the fileHistory was not stored as a global. Maybe it can be an object that gets created to load or save the values, but is not stored otherwise?

aaxu: Hmm yeah, I was just trying to get it to work on one file first and then add other features later on, but I think I can make this work pretty easily.
Right now, how do you open new files? I think we should have tabs to show what file we're on and a button / or shortcut that lets us open a new file.

aaxu: Are all views and objects recreated when you open a new file? What if I made the fileHistory an attribute of the text buffer class? It's pretty much only used for initializing values upon loading anyway.

aaxu:
Moved restoring of the history to move to fileLoad. Note that I had to change the buffer_manger.loadTextBuffer function to set the view of the textBuffer early so that I would be able to initialize the value of scrollView. I had to add the view as an argument for that function so I could access it.

app/history.py
```diff
@@ -19,44 +19,81 @@
...
+    fileSize = os.stat(filePath).st_size
+    return (checksum, fileSize)
+  except:
+    return (None, 0)
+
+def calculateChecksum(filePath):
+  app.log.info("Calculate checksum of the current file")
+  hasher = hashlib.sha512()
+  with open(filePath, 'rb') as file:
+    hasher.update(file.read())
```
dschuyler: Would it be reasonable to use the file data read here:
https://github.com/google/ci_edit/blob/master/app/actions.py#L542
to avoid reading the file twice?

aaxu: Ahh I guess I reverted the change. I wrote this part up because it would allow you to read the file in chunks so you wouldn't have to load the whole thing into memory which would cause an issue if the file was too big. However, I guess that won't really be an issue since this is just text files. I'll change it for now and assume it can read the entire file until we actually hit an issue.

app/actions.py
```diff
@@ -550,30 +550,75 @@ def fileLoad(self):
       self.dataToLines()
     else:
       self.parser = None
+    app.history.loadUserHistory(self.fullPath)
```
dschuyler: Can this get large? Should we be loading it on each text file load?

app/history.py
```diff
@@ -19,44 +19,116 @@
...
+  app.log.info("Calculate checksum of the current file")
+  hasher = hashlib.sha512()
+  if data:
+    hasher.update(data)
+  else:
+    with open(filePath, 'rb') as file:
+      hasher.update(file.read())
+  return hasher.hexdigest()
+
+def clearUserHistory():
```
dschuyler: It this unused (am I just missing where it's used)?

app/buffer_manager.py
```diff
@@ -94,6 +91,7 @@ def loadTextBuffer(self, relPath):
         app.log.info('creating a new file at\n ', fullPath)
       textBuffer = app.text_buffer.TextBuffer()
       self.renameBuffer(textBuffer, fullPath)
+      textBuffer.view = view
```
dschuyler: I still have hope for allowing multiple views to access a textBuffer at the same time. So setting a particular view on a textBuffer will conflict with that. It wasn't enough of an issue to give up the rest of this PR. I just wanted to mention this bit about someday having a many to one relationship between views and a textBuffer.

PR #39, opened 2017-08-08
Bug fix/resizing bugs

### LINKED ISSUE #37: Resizing terminal after starting the program
This just causes a whole bunch of errors. I'll try to list out each one.

* Drag the window so that you make the **height** smaller. Then hold enter and the program will break once the cursor goes past the terminal screen.
* Drag the window so that the **width** is smaller. This will cause a grey area to expand and block the text that is on the screen.
* Increase the size of the window. The size of the **window objects** (like the text buffer, line numbers, etc) will not increase with the size of the window.

aaxu:
Resolves #37.

I commented out that section of code because it was causing a bug. It doesn't seem to affect much when I was playing around with it commented out. Could you explain the purpose of this section @dschuyler ? I was trying to follow it and sort of understand what it was doing, but couldn't figure out what it was trying to accomplish.

### DIFF
--- a/app/ci_program.py
+++ b/app/ci_program.py
@@ -217,7 +217,7 @@ class CiProgram:
 
   def layout(self):
     """Arrange the debug, log, and input windows."""
-    rows, cols = self.cursesScreen.getmaxyx()
+    rows, cols = self.rows, self.cols
     #app.log.detail('layout', rows, cols)
     if self.showLogWindow:
       inputWidth = min(88, cols)
--- a/app/ci_program.py
+++ b/app/ci_program.py
@@ -224,7 +224,7 @@ class CiProgram:
 
   def layout(self):
     """Arrange the debug, log, and input windows."""
-    rows, cols = self.rows, self.cols
+    rows, cols = self.cursesScreen.getmaxyx()
     #app.log.detail('layout', rows, cols)
     if self.showLogWindow:
       inputWidth = min(80, cols)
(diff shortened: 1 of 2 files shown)


dschuyler:
The TopInfo can (in theory) provide N lines of context. e.g.

/file/path.py
class Foo:
  def bar(self):
    if self.value:

would be four lines. The number of lines shown can change and the user moves through the document (so if the scroll position goes above the "if self.value:" it will be in the text area and removed from the context in the TopInfo.

So far, this dynamic updating is imperfect (err, it's annoying actually -- I've been thinking on how to make it less flashy-distracting). So the current workaround is to show two lines (file path, plus a line from the file) all the time. This provides context if the user choose to glance up there, but doesn't flash and jump around.

I should revisit this code and this idea (context lines).

aaxu:
ohh okay that makes a lot more sense. Not sure if it'll be weird if the TopInfo repeatedly expands and contracts though. Can't wait to try this one out :+1:

aaxu:
Just added a check so that the program doesn't crash for now. Will probably need to redo revisit this topinfo feature and make sure that lineCuror < the number of lines when we try to access a line to prevent an index error.

Note: When you keep decreasing the window's height, when the number of visible text lines becomes 0, scrollView becomes 1. This becomes a problem if there is only 1 line of text since we try to retrieve tb.lines[1]

PR #43, opened 2017-08-18
Made moving scroll view smarter

aaxu:
Logic behind this feature:
* If the program opens up and the stored scroll view keeps the cursor in view, keep the same scroll view
* If the cursor is not in view, then search up the values in the preferences file and place the view based on those values. The **scrollRow** will always be based on these value. The **scrollCol** will try to be set to 0 and see if the cursor will be in view. If not, then it will also be based on these values.

This logic also happens when you try to find a string and the scroll view jumps (Implemented in **selectText**).


app/actions.py
```diff
@@ -560,37 +560,64 @@ def restoreUserHistory(self):
...
+      the view should be placed.
+    """
+    optimalRowRatio = app.prefs.editor['optimalCursorRow']
+    optimalColRatio = app.prefs.editor['optimalCursorCol'] 
+    maxRows = self.view.rows
+    maxCols = self.view.cols
+    scrollRow = self.view.scrollRow
+    scrollCol = self.view.scrollCol
+    if not (scrollRow <= row <= scrollRow + maxRows and
+        scrollCol <= col <= scrollCol + maxCols):
```
dschuyler: Whenever I see two <=  (or two >=) on the same statement I wonder if either should be < (or >). I'm guessing that the second <= on each of these should be <, wdyt?

app/actions.py
```diff
@@ -560,37 +560,64 @@ def restoreUserHistory(self):
...
+      # Use optimal position preferences set in default_prefs.py.
```
dschuyler: `# or $HOME/.ci_edit/prefs/editor.py.`

app/actions.py
aaxu: This if statement is supposed to check if the cursor is inside the view. Imagine when **scrollRow** is at 0, and **maxRows** = 25. Then row can be at most 25 since the 25th row is still in the view. The same argument goes for scrollCol, so I think they're both <=. The reason I don't just do > is because I think it improves the understanding of the code if it's written like this.

app/actions.py
aaxu: Ah okay. Also, what's your take on having some way of editing preferences through the program, maybe through another window or some subview?

dschuyler: Yes, I'd like to have the prefs easily editable. I've been thinking of some way to specify the data types and ranges to have a UI built on the fly from that data. Let's start a separate enhancement bug for the topic.

app/actions.py
dschuyler: Confusingly, I admit, maxRows = 25 will show line 1 to line 25 inclusive, but will show row 0 to 24 inclusive. I.e. line numbers are 1 based and rows (and scrollRow) are zero based. Another way to think of it is that all indexes used in the code (not rendered in the UI) is zero based; while the line number, column number shown in the UI are 1 based.

aaxu: Ahh you're right. Forgot about that offset. Done.

app/actions.py
aaxu: Also, does such a file exist? I don't have a $HOME/.ci_edit/prefs/editor.py.

PR #47, opened 2017-08-22
Fixed bug that occurred when default 'python' refers to python3

aaxu:
Changed the she-bang so that it should now refer to python 2 (since we don't have python3 support as of now)

### DIFF
--- a/ci.py
+++ b/ci.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python2
 # TODO(dschuyler): !/usr/bin/python -O
 
 # Copyright 2016 Google Inc.
--- a/ci.py
+++ b/ci.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/python2.7
 # TODO(dschuyler): !/usr/bin/python -O
 
 # Copyright 2016 Google Inc.


ci.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python2
```
dschuyler: On Mac OSX this generates:
/usr/bin/python2: bad interpreter: No such file or directory
(sadly)

aaxu: Hmm how does python get referenced in mac then? I don't own a mac so I can't implement that part well. At best, I can maybe detect the OS type and use python if it's OS and python2 if Linux.

dschuyler:
python2.7 does exist on OSX. Is that available on other platforms?

dschuyler:
Maybe the install.sh could set up the #! line after determining which command to run?

Or the execution could start with a shell script rather than a python script. (essentially doing that executable determination on every startup rather than at install time -- this would only be ok if it were very fast to do).

aaxu:
Ohh yeah on both of my systems, I have both python2.7 and python2 which are two different files (/usr/bin/python2.7 and /usr/bin/python2). They both start up the same python version for me. I'm not sure if this is the case for all the systems, but if that works, then this could be a solution.

PR #51, opened 2017-08-23
update install.sh which can be used under osx

Crabime:
run under osx is ok, maybe i need some modification to be fit for centos later.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -29,14 +29,12 @@ if [ "$(id -u)" != "0" ]; then
     [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
 fi
 
-# Split path to this install script into an array.
-IFS='/' SRC_PATH=($0)
-# Remove this file name.
-unset SRC_PATH[-1]
+#Create opt folder under root
+mkdir -p /opt
+#Get current shell script directory name
+SRC_PATH=$(cd "$(dirname "$0")"; pwd)
 # Grab the directory name.
-APP_DIR="${SRC_PATH[-1]}"
-# Joint path back together.
-SRC_PATH="${SRC_PATH[*]}"
+APP_DIR=${PWD##*/}
 BIN_PATH="/usr/local/bin"
 APP_PATH="/opt"
 INSTALL_DIR="${APP_PATH}/${APP_DIR}"


Crabime:
I signed it!

2017-08-23 9:39 GMT+08:00 googlebot <notifications@github.com>:

install.sh
```diff
@@ -29,14 +29,12 @@ if [ "$(id -u)" != "0" ]; then
...
-unset SRC_PATH[-1]
+#Create opt folder under root
+mkdir -p /opt
+#Get current shell script directory name
+SRC_PATH=$(cd "$(dirname "$0")"; pwd)
 # Grab the directory name.
-APP_DIR="${SRC_PATH[-1]}"
-# Joint path back together.
-SRC_PATH="${SRC_PATH[*]}"
+APP_DIR=${PWD##*/}
```
dschuyler: This looks like it will set the APP_DIR to the directory the script was executed from (rather than the directory with the install.sh script). Those are the same if the script is executed from the same directory as the script.

A fix may be to move the 
```cd "$(dirname "$0")"```
out out of line 35 (maybe put it on its own line above line 35).
That will change the PWD that is used in line 37.

File: install.sh [hunk 2]
```diff
@@ -29,14 +29,12 @@ if [ "$(id -u)" != "0" ]; then
     [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
 fi
 
-# Split path to this install script into an array.
-IFS='/' SRC_PATH=($0)
-# Remove this file name.
-unset SRC_PATH[-1]
+#Create opt folder under root
+mkdir -p /opt
```
johncoleman83: Good catch here

install.sh
```diff
@@ -29,14 +29,12 @@ if [ "$(id -u)" != "0" ]; then
     [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
 fi
 
-# Split path to this install script into an array.
-IFS='/' SRC_PATH=($0)
```
dschuyler: What error occurred with the array?

install.sh
```diff
@@ -29,14 +29,12 @@ if [ "$(id -u)" != "0" ]; then
...
-# Remove this file name.
-unset SRC_PATH[-1]
+#Create opt folder under root
+mkdir -p /opt
+#Get current shell script directory name
+SRC_PATH=$(cd "$(dirname "$0")"; pwd)
```
johncoleman83: I made a PR to use `ci_edit` (or other name of current directory) instead of `.` as the install directory since variable `$0` will have the relative path of `./` with normal install script execution.

install.sh
johncoleman83: I explained more about the error in the other PR #53 ... `$0` = `./install.sh` so the install directory becomes `./`, the other error is in the event that `/opt/` doesn't exist `mkdir -p /opt/` handles that

install.sh
Crabime: thanks

dschuyler:
@Crabime Thanks for working on this. It looks like a CL from @johncoleman83 will change the installer again soon. I'm happy to bring in this CL first if you'd like to touch it up a bit (comments above). It's somewhat ceremonial if it's going to change again right after, but I'm happy to include your contribution CL if you'd like.

Issue #52, opened 2017-08-23
file can not save in osx


Crabime:
Here i changed a file under /opt folder and save it with **control + s**, but it's useless. maybe it's a protected file but if there are some hint about such unwritten condition will be nice. and **cmd + s** will store command line text. i made a screen snapshot but it seems github issue can not upload image..

dschuyler:
Thanks for the bug report!

I've made a change so that the editor will realize the file didn't save. If there are unsaved changed, the editor will prompt the user to about whether they are ok loosing the changes. Does this change help?

Though that prompt may be a good place to suggest save-as; which is rather buried under ctrl+o, ctrl+s

PR #53, opened 2017-08-23
updated install directory to use subdirectory of "ci_edit"

johncoleman83:
@dschuyler, I like the changes you made to safe guard the `rm -rf` command, and all the comments you added.  Also, nice work putting in the `$INSTALL_DIR` in the setup instructions...  I wonder, did you intend to make the install script copy all contents of `ci_edit` to directory `/opt/./`?  When the install script is run like `./install.sh`, `$0` is = to`./install.sh` and since this is a relative path, the method of creating an array of this path ends up making the install directory = `/opt/./`, which becomes a bit redundant and also doesn't organize the package neatly into it's own sub directory (**EDIT**: also this messes with the editors commands -> i.e. ctrl q to quit, no longer functions in the editor -- I don't know why?)...  Originally, it was setup to copy the application to the directory: `/opt/ci_edit/`, which is a bit cleaner as ci_edit would have it's own separate directory for it's package.  Generally `/opt/` is supposed to have the packages in their own subdirectory.

In this PR:

* The assurance that the name of the working directory (and install subdirectory) is not `'./'` by finding the full path of the current directory, `$(basename $(pwd))` (i.e. `$ basename $(pwd)`), then this ensures that the install dir is it's own subdirectory of `/opt/`
* updated documentation and a warning of the usage of `rm -rf`
* I tested in both Linux and Mac Terminal and functions great.
* By the way, Your safeguard check did not continue with the install in my mac terminal when the install directory was `'./'`, so nice work with that.
* Also incorporated PR from @Crabime #52, to ensure that the directory `/opt/` does exist by creating it if it doesn't (`mkdir -p /opt/`).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -40,7 +40,7 @@ fi
 # grab current directory name to use as subdirectory for the installed app
 APP_DIR=$(basename "$SRC_PATH")
 # path for sym link to executable ci.py
-IN_PATH="/usr/local/bin"
+BIN_PATH="/usr/local/bin"
 # location of install directory
 APP_PATH="/opt"
 # full install directory
--- a/install.sh
+++ b/install.sh
@@ -29,24 +29,31 @@ if [ "$(id -u)" != "0" ]; then
     [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
 fi
 
-# Split path to this install script into an array.
-IFS='/' SRC_PATH=($0)
-# Remove this file name.
-unset SRC_PATH[-1]
-# Grab the directory name.
-APP_DIR="${SRC_PATH[-1]}"
-# Joint path back together.
-SRC_PATH="${SRC_PATH[*]}"
+# source path to find current directory and to copy application to /opt/
+SRC_PATH=$(dirname "$0") # path from which install was executed
+SRC_PATH=$(cd "$SRC_PATH" && pwd)  # absolute path
+if [ -z "$SRC_PATH" ]; then
+    # exit if for some reason, the path is empty string
+    [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
+fi
+
+# grab current directory name to use as subdirectory for the installed app
+APP_DIR=$(basename "$SRC_PATH")
+# path for sym link to executable ci.py
 BIN_PATH="/usr/local/bin"
+# location of install directory
 APP_PATH="/opt"
+# full install directory
 INSTALL_DIR="${APP_PATH}/${APP_DIR}"
 
 echo "* *********************************************************** *"
 echo "*                                                             *"
 echo "*   installation steps:                                       *"
 echo "*                                                             *"
-echo "*   (1) copies (or updates) ci_edit directory to the path:    *"
-echo "*       '$INSTALL_DIR'"
+echo "*   (1) copies (or updates) '$APP_DIR' directory to the path: *"
+echo "*       '$INSTALL_DIR'                                        *"
+echo "*       NOTE: update will use the command:                    *"
+echo "*       $ rm -rf $INSTALL_DIR                                 *"
 echo "*                                                             *"
 echo "*   (2) creates (or updates) the execution command:           *"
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -29,14 +29,10 @@ if [ "$(id -u)" != "0" ]; then
...
-# Split path to this install script into an array.
-IFS='/' SRC_PATH=($0)
-# Remove this file name.
-unset SRC_PATH[-1]
-# Grab the directory name.
-APP_DIR="${SRC_PATH[-1]}"
-# Joint path back together.
-SRC_PATH="${SRC_PATH[*]}"
+# source path to find current directory and to copy application to /opt/
+SRC_PATH=$(pwd)
```
dschuyler: Will this require cd to dir with install.sh? I'd like to be able to run the install from elsewhere. like:
sudo stuff/ci_edit/install.sh
Maybe cd within the script itself (similar to what PR from crabime does)?

dschuyler:
My breaking the install from the ci_edit dir is a mistake. It would be awesome to be able to install from ci_edit dir or from outside that dir (which I was adding, and breaking the other).

install.sh
johncoleman83: This did require the `install.sh` to be run in the package directory.  However, I just amended the latest commit, and now you can execute the install script from any directory, and the update will successfully install to `/opt/` using a subdirectory as the package directory.

johncoleman83:
@dschuyler, hey thanks for the feedback.
* I made the install script executable from any directory.  There are comments in that section (lines 33 - 34), which have some explanations.

* I also updated the `-ne` and `-eq` flags for the if conditionals since the comparison was with strings.  Now, lines 89 and 95 use `=` and `! ... =`This is because I was getting a syntax error: 
```
./install.sh: line 89: [[: /opt/ci_edit: syntax error: operand expected (error token is "/opt/ci_edit")
./install.sh: line 94: [[: /opt/ci_edit: syntax error: operand expected (error token is "/opt/ci_edit")
```
Plus, [shellcheck](https://github.com/koalaman/shellcheck/wiki/SC2130) linter for bash scripts throws this error:
```
SC2130: -eq is for integer comparisons. Use = instead.
```

Everything looks good to go now, with the updates for safeguards for not deleting the wrong directory and ensuring all the necessary directories exist; also, the code has been successfully tested to be able to update your installation in the event that there is a previous install, executing the install script from anywhere, and using a custom execution command.

install.sh
```diff
@@ -82,15 +89,19 @@ else
...
-if [[ -z "$APP_PATH" || -z "${APP_DIR}" || "$INSTALL_DIR" -eq "/" ]]; then
+if [[ -z "$APP_PATH" || -z "${APP_DIR}" || "$INSTALL_DIR" = "/" ]]; then
   echo "Something is incorrect about the install directory. Exiting."
   exit -1
 fi
 # Yes, this is redundant with the above. User safety is top priority.
-if [[ -n "$APP_PATH" && -n "${APP_DIR}" && "$INSTALL_DIR" -ne "/" ]]; then
-  rm -rf "$INSTALL_DIR"
+# Only continues to delete if directory exists and is not root '/'
+if [[ -n "$APP_PATH" && -n "${APP_DIR}" && ! "$INSTALL_DIR" = "/" ]]; then
```
dschuyler: Thanks! That note about -eq is good to know. After reading up on it more it looks like this is now POSIX compliant, which is cool. Though the = in a conditional stings my c++/python mind :)
Can != be used here and == used on line 92? (I admit, just for my own sanity). wdyt? (and is the shell checker cool with != and ==)

johncoleman83: I updated to `==` and `!=` as it is absolutely more common across languages and therefore less confusing.  I think it's good to have readable code, and especially since this passes the `shellcheck` linter.

johncoleman83:
p.s. If the PR from @crabime is incorporated, I'll be sure to rebase and fix any merge conflicts if necessary.

dschuyler:
Thanks!

PR #55, opened 2017-08-23
Made all labels have the message_line color

### LINKED ISSUE #54: LabeledLine color is wrong
I'm not sure if it was intentional, but the labels in labeled line didn't seem to have the right color IMO. They were using the default color, which is based on the color of the background of the terminal you are using. Thus, for the Ubuntu default terminal, the window would look something like this:

image

aaxu:
Now all messages that appear in LabeledLine will have the yellowish color. This color differs from terminal to terminal but is all sort of yellow (from what I found). I could use the default color (black text on white background), but this seems to catch the eye more so the user can easily see that some text just appeared on the screen. Resolves #54

### DIFF
--- a/app/window.py
+++ b/app/window.py
@@ -280,7 +280,7 @@ class LabeledLine(Window):
 
   def refresh(self):
     self.leftColumn.addStr(0, 0, self.label,
-        app.prefs.color['default'])
+        app.color.get('message_line'))
     Window.refresh(self)
 
   def reshape(self, rows, cols, top, left):


dschuyler:
My version of that color must be different I'm guessing. I see dull yellow text on a white background (which doesn't pop, it's very muted).
image

aaxu:
Hmm yeah that's true. On my Ubuntu terminal, it was too bright and didn't have much contrast with the background. Maybe something darker like blue or brown may work better. I'll send another PR and you can see if it works well on your terminals.

PR #56, opened 2017-08-23
New color for status line

aaxu:
I figured a bright blue would work. It seems to be somewhat more consistent throughout the terminals. See if this is true for you as well. I used the keyword color, since it's currently not being used by much. If we end up using this color more in the text, we may have to change this status color since it won't be the only blue anymore and won't pop out as much.

I could always change the background part of it too and make it unique, but I think that's less aesthetically pleasing than if the background matched its surroundings

image

### DIFF
--- a/app/window.py
+++ b/app/window.py
@@ -280,7 +280,7 @@ class LabeledLine(Window):
 
   def refresh(self):
     self.leftColumn.addStr(0, 0, self.label,
-        app.color.get('message_line'))
+        app.color.get('keyword'))
     Window.refresh(self)
 
   def reshape(self, rows, cols, top, left):


aaxu:
Actually changed it to avoid conflicts. What do you think about this color? Is it too light? It's not used for anything right now and can catch the eye, but not sure if it has too little contrast with the white background to make it hard to see.

image

aaxu:
Oh okay, I thought it was going to be used for something since it was named 'keywords'. For example, if all keywords like **if, and, or** are highlighted blue, then if your code populates the view with blue, the labeled line won't pop up as much since it's of the same color. However, it seems as though nothing is using it right now so I can revert the change back.

aaxu:
Reverted back to previous commit with the blue text.

PR #60, opened 2017-08-30
Enhancement/bookmarks

aaxu:
Bookmarks should be working as intended now.

## Bookmark Rules:
* Setting a bookmark will save the cursor's position as well as any highlighted text.
* If you try to set any part of a bookmark (for example, if a bookmark saved was originally 5 lines tall and you use the remove bookmark command on any one of those lines, the entire bookmark will be deleted.
* If you try a part of a bookmark over any previously existing part of a bookmark, the existing one will be removed and the new one will be added
* Jumping to a bookmark will jump to the one closest to the cursor's location (forward for F2 and backward for SHIFT + F2).

When you jump to a bookmark, it currently does not use optimal scroll position. I have a PR coming for that later on that should support optimal cursor position for highlighted block of text, but I still have some questions that need to be answered first before I can finish them (I made an issue for them).


app/actions.py
```diff
@@ -92,30 +92,172 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
+    Args:
+      None
+
+    Returns:
+      None
+    """
+    while len(self.bookmarks) < len(self.lines):
+      # Double the length of the array
+      length = len(self.bookmarks)
+      self.bookmarks.extend([None] * max(1, length))
```
dschuyler: I'm still looking over this PR, so I'm not sure why this is looking for the next higher power of two value, but I think this line may give the desired number without looping:
```
newBookmarksLength = 1<<(len(self.lines)-1).bit_length()
```

app/actions.py
```diff
@@ -92,30 +92,172 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
+    its length is at least the number of lines in the document.
+
+    Args:
+      None
+
+    Returns:
+      None
+    """
+    while len(self.bookmarks) < len(self.lines):
+      # Double the length of the array
```
dschuyler: I'm concerned about having another array that is as large as the document has lines. Some documents I work with have 20000 lines or more. :) 

Maybe I'm different, but I don't tend to use more than a half dozen bookmarks (just enough to work through a problem).

I'm wondering if this PR is thinking in different scales. Like 2000 lines and hundreds of bookmarks(?).

app/actions.py
aaxu: Oh wow I didn't know there was such a function. Good to know!

app/actions.py
aaxu: Yeah I was kind of thinking about all the cases. Even though most people only use a few, we don't want the program to crash or have bookmarks overwrite other ones silently. I was also trying to make it quick because although it does take more memory, without this list, you would have to search through a list of sets. However, let me try try this other implementation since I don't think it'll pose a major performance issue.

aaxu:
Updated bookmarks to use less space.

app/actions.py
```diff
@@ -92,30 +92,170 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
     app.log.debug()
-    self.view.bookmarkIndex -= 1
-    self.bookmarkGoto()
+    bookmark = None
+    upperRow, _, _, _ = self.startAndEnd()
+    numBookmarks = len(self.bookmarks)
+    tempBookmark = ((upperRow,),)
+    index = bisect.bisect_left(self.bookmarks, tempBookmark)
+    if numBookmarks:
+      bookmark =  self.bookmarkGoto(self.bookmarks[index - 1])
```
dschuyler: There's an extra space after the =
It doesn't look like bookmark is used.

app/actions.py
```diff
@@ -92,30 +92,170 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
+    else:
+      self.setMessage("No bookmarks to jump to")
```
dschuyler: It looks like numBookmarks is only used for this if. Please instead, move this up to the top of this function:
```
if not len(self.bookmarks):
  self.setMessage("No bookmarks to jump to")
  return
```
i.e. early out.
And remove |numBookmarks|

app/actions.py
```diff
@@ -92,30 +92,170 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
+    """
     app.log.debug()
-    self.view.bookmarkIndex -= 1
-    self.bookmarkGoto()
+    bookmark = None
+    upperRow, _, _, _ = self.startAndEnd()
+    numBookmarks = len(self.bookmarks)
+    tempBookmark = ((upperRow,),)
+    index = bisect.bisect_left(self.bookmarks, tempBookmark)
+    if numBookmarks:
```
dschuyler: Was this intending to check  |index|?

app/text_buffer.py
```diff
@@ -31,6 +31,8 @@ def __init__(self):
     self.fileHistory = {}
     self.lastChecksum = None
     self.lastFileSize = 0
+    self.bookmarks = []
+    self.bookmarkSets = []
```
dschuyler: Maybe remove line 35? I don't see self.bookmarkSets being used.

app/actions.py
```diff
@@ -92,30 +92,170 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
 
   def bookmarkRemove(self):
+    """
+    Removes bookmarks in all selected lines.
+
+    Args:
+      None.
+
+    Returns:
+      None.
```
dschuyler: What do ya think about leaving out Args and Returns if there aren't any to speak of? I'm flexible on it if ya really like it this way.

app/actions.py
```diff
@@ -92,30 +92,170 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
-    self.bookmarkGoto()
+    bookmark = None
+    _, _, lowerRow, _ = self.startAndEnd()
+    numBookmarks = len(self.bookmarks)
+    tempBookmark = ((lowerRow,float('inf')),)
+    index = bisect.bisect(self.bookmarks, tempBookmark)
+    if numBookmarks:
+      bookmark =  self.bookmarkGoto(self.bookmarks[index % numBookmarks])
+    else:
+      self.setMessage("No bookmarks to jump to")
```
dschuyler: Please see notes on bookmarkPrior (yeah, I read that function first).

aaxu:
@dschuyler  Hey, just wondering if there was anything else about these PRs? Bookmarks should use a little amount of space now.

app/actions.py
```diff
@@ -92,30 +92,170 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
-    bookmark = app.bookmarks.get(self.view.bookmarkIndex)
-    if bookmark:
-      self.selectText(bookmark['row'], bookmark['col'], bookmark['length'],
-          bookmark['mode'])
+    bookmarkRange, bookmarkData = bookmark
+    cursorRow, cursorCol = bookmarkData['cursor']
+    penRow, penCol = bookmarkData['pen']
+    markerRow, markerCol = bookmarkData['marker']
+    selectionMode = bookmarkData['selectionMode']
+    self.__skipCursorScroll = True
```
dschuyler: This might be something from another PR. Let's separate it from this PR. i.e. delete line 172.

app/actions.py
```diff
@@ -92,30 +92,170 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
+      bookmarkRange2 (tuple): a sorted tuple of row numbers.
+
+    Returns:
+      True if the ranges overlap. Otherwise, returns False.
+    """
+    if (bookmarkRange1[-1] >= bookmarkRange2[0] and
+        bookmarkRange1[0] <= bookmarkRange2[-1]):
+      return True
+    else:
+      return False
```
dschuyler: Please do instead
```
return (bookmarkRange1[-1] >= bookmarkRange2[0] and
       bookmarkRange1[0] <= bookmarkRange2[-1])
```

app/actions.py
```diff
@@ -92,30 +92,170 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
+    """
+    cursor = (self.view.cursorRow, self.view.cursorCol)
+    pen = (self.penRow, self.penCol)
+    marker = (self.markerRow, self.markerCol)
+    selectionMode = self.selectionMode
+    bookmarkData = {
+      'cursor': cursor,
+      'marker': marker,
+      'pen': pen,
+      'selectionMode': selectionMode,
```
dschuyler: I like use temps when the lines are too long. In this case, it looks like the values will fit nicely (i.e. readably) directly in lines 113 to 116. WDYT?

app/actions.py
```diff
@@ -92,30 +92,170 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
 
-  def bookmarkGoto(self):
+    Args:
+      None.
+
+    Returns:
+      None.
+    """
+    newBookmark = self.dataToBookmark()
+    self.bookmarkRemove()
```
dschuyler: Can two bookmarks overlap? I'm okay either way I think. If they cannot overlap, I think the remove code could be simplified and maybe we can get by with line 119 being something like
```
bookmarkRange = (upperRow, lowerRow + 1)
```
because we'd know that if we were in those bounds then then we found the only bookmark that needs to be removed.

(I'm also a bit concerned for someone doing select all, add bookmark and having |bookmarkRange| be a very large tuple).

app/actions.py
aaxu: Sorry, I'm not really understanding what you mean. What do you mean by using temps and what values are you suggesting to inject?

app/actions.py
aaxu: Done

app/actions.py
aaxu: Ah you're right. This was left over from previous implementation. Changed!

app/actions.py
aaxu: woops :D got too excited working on it

app/actions.py
aaxu: Yeah, so self.bookmarks is a sorted list of ( (range), bookmarkDataDict). I'm using tempBookmark be assigned to that so that bisect.bisect_left gets the proper index of where it should be. Once I knew that, I could find the bookmark that would precede it and that is where we would jump to.

app/actions.py
aaxu: woops! it doesn't even return anything either haha

app/actions.py
aaxu: :+1:

app/actions.py
aaxu: Uhh I'm still kind of experimenting and learning documentation styles. I thought about leaving it out, but thought maybe it would be more clear for skimming through functions as a reminder of what they take/return. It does take up a lot of space, but it's what I saw on Google's Python style docs so I figured I would give it a try. 

I think we could keep it to keep the format consistent, or else we should at least document it in the docstring comment to detail what the function's arguments/return values are (so someone doesn't think it's just incomplete documentation).

app/text_buffer.py
aaxu: Woops! Old code!

app/actions.py
dschuyler: Something like this
```
  bookmarkData = {
    'cursor': (self.view.cursorRow, self.view.cursorCol),
    'marker': (self.markerRow, self.markerCol),
    'pen': (self.penRow, self.penCol),
    'selectionMode': self.selectionMode,
  }
```

aaxu: Ahh I see. Done.

app/actions.py
aaxu: Edit: bookmarks cannot overlap

aaxu:
All fixed!

app/actions.py
```diff
@@ -92,30 +92,158 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
+
     app.log.debug()
-    return app.bookmarks.remove(self.view.bookmarkIndex)
+    upperRow, _, lowerRow, _ = self.startAndEnd()
+    numBookmarks = len(self.bookmarks)
+    bookmarkRange = (upperRow, lowerRow)
+    if self.bookmarks:
+      helpRemoveBookmarks(bookmarkRange, 0, numBookmarks - 1)
+      self.bookmarks = [b for b in self.bookmarks if b]
+    return len(self.bookmarks) != numBookmarks
```
dschuyler: What do you think of changing this from a recursive sub-devision into a lookup and a linear walk?

Here's a hackish example
```
a = [(0, 5), (8, 10), (12, 12), (15, 20)]
tests = [(0, 0), (0, 10), (7, 10), (8, 20), (30, 30), (0, 500)]
print 'test list', a

def removeOverlaps(rangeList, needle):
  index = bisect.bisect_left(a, needle)
  if index and needle[0] <= rangeList[index-1][1]:
    index -= 1
  begin = end = index
  while True:
    index += 1
    end = index
    if index >= len(rangeList) or needle[1] < rangeList[index][0]:
      break
  return rangeList[:begin] + rangeList[end:]

for r in tests:
  print 'range', r, 'after remove', removeOverlaps(a, r)
```

app/actions.py
```diff
@@ -92,30 +92,158 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
+
```
dschuyler: [Deleted suggestion for a change. My suggestion was flawed].

dschuyler:
I may tinker with bookmarkRemove a bit. I'll try not to break it :)

PR #61, opened 2017-08-31
Bug fix/shift page up down

aaxu:
It looks to be fixed! Let me know if this works for your systems! :smiley:


app/curses_util.py
```diff
@@ -68,7 +68,9 @@
 KEY_HOME = curses.KEY_HOME
 KEY_END = curses.KEY_END
 KEY_PAGE_DOWN = curses.KEY_NPAGE
+KEY_SHIFT_PAGE_DOWN = curses.KEY_SNEXT
 KEY_PAGE_UP = curses.KEY_PPAGE
+KEY_SHIFT_PAGE_UP = curses.KEY_SPREVIOUS
```
dschuyler: On Mac (iterm2 and Terminal) and ubuntu (whatever the default terminal in cinnamon is) the shift + page up/down is caught by the terminal software to do page up/down in the terminal history :( I get either no key input or KEY_NPAGE (for shift + page down; and page down; i.e. the same key code is sent).

app/actions.py
```diff
@@ -401,7 +412,18 @@ def cursorPageDown(self):
         self.cursorColDelta(self.penRow + penRowDelta), 0, 0)
     self.redo()
 
-  def cursorPageUp(self):
+  def __cursorPageUp(self):
```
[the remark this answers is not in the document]
aaxu: Oh woops!

app/cu_editor.py
```diff
@@ -69,8 +69,10 @@ def mainWindowCommands(controller, textBuffer):
     KEY_ESCAPE: textBuffer.normalize,
     KEY_F1: controller.info,
     KEY_BTAB: textBuffer.unindent,
-    KEY_PAGE_UP: textBuffer.cursorPageUp,
-    KEY_PAGE_DOWN: textBuffer.cursorPageDown,
+    KEY_PAGE_UP: textBuffer.cursorNeutralPageUp,
+    KEY_PAGE_DOWN: textBuffer.cursorNeutralPageDown,
```
dschuyler: Maybe we can do this for now:
```
# TODO(aaxu): Enable neutral vs. Select page up/down once we can
# differentiate those keys on Mac and Ubuntu terminals.
#KEY_PAGE_UP: textBuffer.cursorNeutralPageUp,
#KEY_PAGE_DOWN: textBuffer.cursorNeutralPageDown,
KEY_PAGE_UP: textBuffer.cursorSelectPageUp,
KEY_PAGE_DOWN: textBuffer.cursorSelectPageDown,
```

app/actions.py
```diff
@@ -416,6 +438,76 @@ def cursorPageUp(self):
         self.cursorColDelta(self.penRow + penRowDelta), 0, 0)
     self.redo()
 
+  def cursorNeutralPageDown(self):
+    """
+    Moves the view and cursor down by a page or stops
+    at the bottom of the document if there is less than
+    a page left. Does not select any text and removes all
+    existing highlights.
+    Args:
```
dschuyler: Is there a guideline these comments are following?
Just noticing that there have been blank lines above Args: line. I'm ok either way, just thought ya may want it consistent.

aaxu: Oh yeah my bad LOL this was how i saw it done on google's python style docs so I tried doing it. I'm still not used to it, but there's supposed to be a blank line before that.

app/cu_editor.py
aaxu: I think we should use cursorNeutralPage for the regular keys and just keep  the cursorSelectPageUp/Down out for now. One of the reasons for this is because it was unexpected since I wasn't holding shift. The other reason is because my other PR ( #62 ) would break page down when selection mode is not none since it would keep trying to show the top of the selection.

dschuyler: I see your point. In that case I'm leaning toward keeping this PR with the Neutral and Select versions intact.

app/curses_util.py
dschuyler: Your changes/code are correct. The bug is elsewhere and my code was just trying to compensate for bugs elsewhere.

PR #62, opened 2017-08-31
Enhancement/optimal scroll position

aaxu:
Provides support for snapping back to highlighted text (bookmarks)!

## Changes
* Removed all direct assignments to self.view.scrollRow/Col. This is now all done in **updateScrollPosition** function now. You can control whether a function manually updates the view position by utilizing self.__skipCursorScroll and self.__skipUpdateScroll.
* When you have a highlighted block of text and snap to it, the program will try to set that block of text at a position according to the optimal position preferences in default_prefs.py
* Anything that causes the screen to snap to a position (not scroll up/down) should make use of the optimal cursor position.


aaxu:
I actually wanted to change updateScrollPosition from cursorPageUp/Down to cursorMoveScroll so it would be easy to keep track of what called it for future debugging, but it would be slightly less efficient. It wouldn't really provide any visible decrease in performance. What do you think about this change @dschuyler?

aaxu:
Maybe it would actually be better if we were to just make everything call updateScrollPosition when they want to move the scroll position (thus replacing the purpose of cursorMoveScroll). Any previous call to cursorMoveScroll could then call updateScrollPosition and then cursorMove.

I think maybe I'll work on refactoring/documentation sometime soon (although I don't really know all the functions quite yet).

app/actions.py
```diff
@@ -1270,17 +1292,32 @@ def unindentLines(self):
...
-    maxRow, maxCol = self.view.rows, self.view.cols
-    if self.view.scrollRow > self.penRow:
-      self.view.scrollRow = self.penRow
-    elif self.penRow >= self.view.scrollRow + maxRow:
-      self.view.scrollRow = self.penRow - (maxRow - 1)
-    if self.view.scrollCol > self.penCol:
-      self.view.scrollCol = self.penCol
-    elif self.penCol >= self.view.scrollCol + maxCol:
-      self.view.scrollCol = self.penCol - (maxCol - 1)
+    else:
```
dschuyler: Since this returns, there's no need to make this an else.

app/actions.py
```diff
@@ -38,7 +38,12 @@ def __init__(self):
     app.mutator.Mutator.__init__(self)
     self.view = None
     self.rootGrammar = app.prefs.getGrammar(None)
+    # |__skipUpdateScroll| is False if you do not want the program
+    # to automatically set the view to the optimal scroll position.
     self.__skipUpdateScroll = False
```
dschuyler: If there's an opportunity to remove these __skip* flags, please go for it. I'd prefer not to have them, but I understand it can be hard to avoid at times.

aaxu:
Updated with new changes. Requested changes have been processed. Lots of refactoring and dividing functions into smaller ones!

app/actions.py
aaxu: I actually think we do need them just because our updateScrollPosition function is called upon refreshing the screen, and sometimes we don't want to update the scroll position (specifically for if captive cursor is false). I refactored some functions, so now the __useOptimalScroll flag tells the function if it should use the basic or optimal scroll position algorithm. I don't see a real easy way to get rid of these flags for now.

PR #65, opened 2017-09-01
Bug fix for file encoding

### LINKED ISSUE #64: Saving a new file
This causes problems and can break future openings of ci_edit if you open up a new file and try to save. This is because self.fileEncoding is not initialized when you open a new file.

aaxu:
Resolves #64


dschuyler:
Thanks

PR #66, opened 2017-09-01
Bug fix for changing selection mode on select page up/down

aaxu:
For example, if you are in kSelectionBlock which highlights columns 1-5 and you use SHIFT+PAGEDOWN, only columns 1-5 in all those rows will be highlighted. If you're in kSelectionLine, all lines will be highlighted (rather than just all text inside those lines).


dschuyler:
I think we both fixed this :) I added a check for "if selection is none" then change to char. If that looks good to you I may close this PR without merging.

aaxu:
Oh yeah I saw your change. However, if selection is kSelectionBlock for example, the selection mode won't change to kSelectionCharacter and it will end up causing weird behaviors. That's why I changed your check to make sure it always becomes kSelectionCharacter if it is in some other mode.

dschuyler:
Really, what is the strange behavior? I tried the select page up/down (by making different keys bound to those actions since I can't seem to shift page up). The selection block appeared to work correctly (but maybe I missed something).

aaxu:
Steps to reproduce error:
1. Go to **cursorSelectPageUp** and change the if block to simply self.selectionBlock() just to test cases for what would happen if selection mode is at kSelectionBlock.
2. type the following in a new file in ci_edit so it looks like the following:
hi there 

 hello there

 hello there friend

app/actions.py
```diff
@@ -487,7 +487,9 @@ def cursorSelectPageDown(self):
     Returns:
       None.
     """
-    if self.selectionMode == app.selectable.kSelectionNone:
+    # Selection mode must be kSelectionCharacter
+    # or else weird behaviors will occur.
+    if self.selectionMode != app.selectable.kSelectionCharacter:
       self.doSelectionMode(app.selectable.kSelectionCharacter)
```
dschuyler: I think the comment is more a note to me in this review, so let's remove it from the code/PR.
Having 
if foo != X:
  foo = X
is the same as writing
foo = X
Please go ahead and remove the 'if' line.

app/actions.py
```diff
@@ -508,7 +510,9 @@ def cursorSelectPageUp(self):
     Returns:
       None.
     """
-    if self.selectionMode == app.selectable.kSelectionNone:
+    # Selection mode must be kSelectionCharacter
+    # or else weird behaviors will occur.
+    if self.selectionMode != app.selectable.kSelectionCharacter:
```
dschuyler: similar notes to above.

app/actions.py
aaxu: I was thinking that by having this check, it would save some time since calling doSelectionMode seems very expensive so the if statement was more of a performance decision. 

I do think, however, that I should change the call to self.selectionCharacter(). 

Do you still want me to delete the if statement?

aaxu:
Can you elaborate more on what we should leave it as? For now, I've implemented the functions for alt+shift+page up/down which will always use selectionBlock. The default page up and page down will use remove all highlights and holding shift will use kSelectionCharacter. 

Tell me if I'm wrong but did you mean that we should just not deal with the selection mode when using page up/down and just let the weird behaviors happen for now?

aaxu:
Also not really sure how to describe the function since it seems to select columns, but only after a certain condition is true, like markerCol is diff from penCol or something like that.

app/actions.py
```diff
@@ -629,10 +629,40 @@ def cursorSelectPageUp(self):
...
+      None.
+
+    Returns:
+      None.
+    """
+    if self.selectionMode != app.selectable.kSelectionBlock:
+      self.selectionBlock()
+    self.__cursorPageDown()
+
+  def cursorAltSelectPageUp(self):
```
dschuyler: how about the name cursorSelectBlockPageUp

app/actions.py
```diff
@@ -629,10 +629,40 @@ def cursorSelectPageUp(self):
...
+    DESCRIPTION HERE
+
+    Args:
+      None.
+
+    Returns:
+      None.
+    """
+    if self.selectionMode != app.selectable.kSelectionBlock:
+      self.selectionBlock()
```
dschuyler: If this call were inside an inner-loop (called frequently) then avoiding the function call can make a big performance difference. 

In this case the inner function has a 'if' to early out if it is being set to the same value. 

I'd prefer not to have the condition for these at this level so the code is cleaner to read.

(I like to have code be cleaner and easier to read, but there are times (lots of times) I will write weird code for the sake of performance. i.e. If a user may notice the difference I'll sacrifice how the code looks. In this case, I don't think a user will be able to notice the difference).

app/actions.py
```diff
@@ -629,10 +629,40 @@ def cursorSelectPageUp(self):
...
+  def cursorAltSelectPageUp(self):
+    """
+    DESCRIPTION HERE
+
+    Args:
+      None.
+
+    Returns:
+      None.
+    """
```
dschuyler: With a name like cursorSelectBlockPageUp() for a function that calls
Selection Block and cursor PageUp, can the doc string be left off.

app/actions.py
```diff
@@ -629,10 +629,40 @@ def cursorSelectPageUp(self):
...
+    if self.selectionMode != app.selectable.kSelectionBlock:
+      self.selectionBlock()
     self.__cursorPageUp()
 
+
+
```
dschuyler: Please use one space between functions.

app/em_editor.py
```diff
@@ -487,6 +487,8 @@ def setTextBuffer(self, textBuffer):
       KEY_PAGE_DOWN: textBuffer.cursorNeutralPageDown,
       KEY_SHIFT_PAGE_UP: textBuffer.cursorSelectPageUp,
       KEY_SHIFT_PAGE_DOWN: textBuffer.cursorSelectPageDown,
+      KEY_ALT_SHIFT_PAGE_UP: textBuffer.cursorAltSelectPageUp,
+      KEY_ALT_SHIFT_PAGE_DOWN: textBuffer.cursorAltSelectPageDown,
 
       CTRL_A: textBuffer.selectionAll,
```
dschuyler: Just chatting: context about removing "Alt" from the function name (in comment above), somethin' I'm going for is to have the key bindings be the keys in this dictionary and the value functions (actions) be named by what happens (with no mention of the key, if possible). 

This is so a different keybinding reads well. e.g. some other keybinding might use F9 to selectionAll, and it would be weird for the function name selectionAll to have Ctrl or A or F9 be part of the action function name.

An example is the CTRL_A: textBuffer.selectionAll, CTRL_A: doesn't say anything about what it does and textBuffer.selectionAll doesn't say anything about what key it's bound to (I think that's good).

dschuyler:
I think this PR is going in the right direction. 

We don't need to keep the weird behavior I'd previously coded.

Yeah, block selection is unusual because it may have no width. e.g. in may go from row 5 to row 10, but only go from col 8 to col 8 (i.e. have no width at all). It makes a difference when when the user extends the select left or right (and that whole height becomes selected). Try it in a GUI editor (it likely will behave the same).

app/actions.py
aaxu: It does have a different behavior though, so we should probably document its function. Just don't know if how kSelectionBlock's implementation is making the effect you want from the function.

app/actions.py
aaxu: Okay sounds good! I'll keep that in mind from now. I've been trying to write clean and easy to read code (like if you read it aloud, you would understand it without having to think too much about like some idoms in python such as return x == y or y == z etc). Still conflicted whether I should use these or not, or make it more clear through if statements. What's your take on this?

app/actions.py
aaxu: Woopsies. Will do

app/em_editor.py
aaxu: Oh okay, I see now. Thanks for the tip. I'll try to keep that in mind.

app/actions.py
aaxu: Sounds good to me.

aaxu:
Wow honestly didn't even know that was a function in text editors LOL good to know.

app/actions.py
aaxu: Edit: it behaves like the GUI editor so may I'll just leave it off then. I'll update the cursorSelectPageUp/Down docs too

aaxu:
Changed cursorNeutralPageUp/Down to cursorSelectNonePageUp/Down for symmetry.
Changed cursorSelectPageUp/Down to cursorSelectCharacterPageUp/Down for symmetry.
Changed documentation to reference page up/down instead of copying the description.

PR #69, opened 2017-09-08
Feature: show the last key combination pressed in status line when the --log flag is present

aaxu:
I thought this would be useful for debugging keys in case anyone else wants to use it.


dschuyler:
This looks like the same info that is shown in the right-hand debug window. Should we consider removing it from the debug window?

aaxu:
omg LOL i never knew that was a feature cause nothing showed up since I didn't use it on full screen a lot . Can ignore this then!

Issue #88, opened 2017-10-10
Undo into saved undo breaks redo chain


aaxu:
Ways to emulate this:
* Copy anything into your clipboard.
* Create a new file.
* Paste consecutively three times
* Save
* Reopen file
* Hold undo
* Weird behaviors will happen. The best behavior I've seen is that the redo chain just stops allowing you to undo, even when there is supposed to be more stuff. (ie, doing the steps above should leave one copy paste in your program). The worst behavior I've seen is where the views all freeze. I think you can still exit with CTRL_Q + y/n, but the visible screen looks like it froze. Another behavior is where the screen looks frozen, but once u click the text buffer, changes get updated as if the screen refreshed.

dschuyler:
I'm able to repro it.
running we --clearHistory will make the issue not happen (though that also removes the history -- this is not a suggestion, just a point of data).
I tried going back dozens of git commits and the issue persist.

dschuyler:
As a stop-gap (not a real solution) I made "saveUndo" a preference setting and default it to False.
So a new user won't see this issue, but they also wont have save/restore undo.

PR #91, opened 2017-10-16
Fix pasting from breaking the redo chain

aaxu:
I think when you removed that line `len(self.redoChain) != self.redoIndex`, it might've caused this bug. The bug was caused because pasting consisted of an `insertLines` part and then a `cursorMove` part. However, I think removing that line caused the `cursorMove` part to become a `tempChange`. tempChange is not saved/restored in user history, so that sometimes caused an issue when reopening the file.

I combined the `insertLines` and the `cursorMove` part into one `paste` change, so this should fix the bug.
There probably could have been other fixes like saving tempChange, but that would mean trivial actions would also get saved and I don't know if that would be useful or annoying. We could also probably redo the logic and keep the paste change as 2 separate parts, but I thought this would be cleaner.

### DIFF
--- a/app/ci_program.py
+++ b/app/ci_program.py
@@ -521,6 +521,9 @@ class CiProgram:
         elif i == '--version':
           userMessage(app.help.docs['version'])
           self.quitNow()
+        elif i == '--clearHistory':
+          app.history.clearUserHistory()
+          self.quitNow()
         elif i.startswith('--'):
           userMessage("unknown command line argument", i)
           self.quitNow()
--- a/app/ci_program.py
+++ b/app/ci_program.py
@@ -592,7 +592,6 @@ class CiProgram:
       self.dirPrefs = os.path.join(homePath, 'prefs')
       if not os.path.isdir(self.dirPrefs):
         os.makedirs(self.dirPrefs)
-      app.history.path = os.path.join(homePath, app.history.path)
     except Exception, e:
       app.log.error('exception in makeHomeDirs')
 
--- a/app/ci_program.py
+++ b/app/ci_program.py
@@ -21,7 +21,6 @@ if os.getenv('CI_EDIT_USE_FAKE_CURSES'):
 
 
 import app.background
-import app.bookmarks
 import app.curses_util
 import app.help
 import app.history
--- a/app/default_prefs.py
+++ b/app/default_prefs.py
@@ -115,7 +115,6 @@ prefs = {
     'c_string2': stringColorIndex,
     'debug_window': defaultColorIndex,
     'regex_string': stringColorIndex,
-    'regex_string2': stringColorIndex,
     'doc_block_comment': commentColorIndex,
     'html_block_comment': commentColorIndex,
     'html_element': keywordsColorIndex,
@@ -421,7 +420,7 @@ prefs = {
       ],
       'contains': [
         'c_string1', 'c_string2', 'doc_block_comment', 'cpp_block_comment',
-        'cpp_line_comment', 'regex_string', 'regex_string2',
+        'cpp_line_comment', 'regex_string',
       ],
     },
     'keyword': {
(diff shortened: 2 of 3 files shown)


aaxu:
Also changed default back to using saveUndos again, since this should resolve #88

aaxu:
Ah yes, those should probably be changed as well. I'll probably see to it tomorrow morning since I have a school project to do all night today :(

aaxu:
I'm actually thinking that  just saving self.tempChange in file history would be better. It would fix all of these issues that occur when an action ends in a cursor move. Do you think we should do that, or manually edit all the changes so they're actually 1 change rather than a sequence of changes?

I think it makes more sense for all the changes (not just ones that end in a cursor move) of an action to only have one element in the redo chain, but I feel like we'd have to do a lot of remodeling and many assumptions and functions would probably break. However, I'm up for it if you're willing. Let me know what you think. Maybe even save temp change as a short term fix for release and work on remodeling the code?

dschuyler:
To experiment with grouping the actions, ya could wrap the calls as compound changes.
```python
    self.compoundChangeBegin()
    [...]
    self.compoundChangeEnd()
```
That will allow make the changes within the compound change undo/redo with a single key combination.

When experimenting, try bundling the (move, action) as well as (action, move) i.e. with the order changed. My concern/thinking is that can be hard to see what is changing when undoing/redoing depending on how the moves are combined.

The intended behavior was to do (move, action) when redoing; and do the reverse (action, move) when undoing. This seems to be what GUI editors do. (though I just tried it and I see that it's not currently working that way i.e. a bug).

If the move is bundled with the action, I think that would make reversing harder, because the 'move' mentioned above is two different moves. e.g. if numbers are actions and letters are moves in this sequence: 1a2b3c4d5e6 and | is the redo marker then the marker would change like this:
```
start: 1a2b3c4d5e6|
 undo: 1a2b3c4d5|e6
 undo: 1a2b3c4|d5e6 
 undo: 1a2b3|c4d5e6
 redo: 1a2b3c|4d5e6
 redo: 1a2b3c4d|5e6
 redo: 1a2b3c4d5e|6
 redo: 1a2b3c4d5e6|
```
See how the first redo only does (c,); and second undo does (5, d) while the second to last redo does (5, e). When undoing 5 is pairs with d and when redoing 5 is paired with 3.

aaxu:
>  try bundling the (move, action) as well as (action, move) i.e. with the order changed. My concern/thinking is that can be hard to see what is changing when undoing/redoing depending on how the moves are combined.

Can you give an example of when this would be a concern? I was thinking that when you redo, you would handle bundles from left to right and when you undo, you would handle them from right to left. This should be the intended behavior that you wanted.

> If the move is bundled with the action, I think that would make reversing harder, because the 'move' mentioned above is two different moves.

If they are bundled, wouldn't it make it easier, since we then know we only have to handle this bundle?

> The first redo only does (c,); and second undo does (5, d) while the second to last redo does (5, e). When undoing 5 is pairs with d and when redoing 5 is paired with e.

In this example, I think this would only happen when we move the cursor with either the arrow keys or the mouse. For example, if the sequence looked like this (parenthesis is a bundle):

```
start: 3a(4b)c|
 undo: 3a|(4b)c
 redo: 3a(4b)|c
```

Here I see that redo and undo is not the same, but it allows the user to see what they undo and redo. If the redo did (4b)c, then the last cursor move could move the view out of sight and the user could be confused about what they 'redo'd. I think this is also the behavior of many other GUI editors, so in this case, it should be fine?

dschuyler:
Hmm, in the example 4bc, that seems to be an unlikely pattern because moves (b and c) are moves that would be combined by the redo chain optimizer. Looking at my redo chains it seems common to alternate between some action (like 'i') and a move (like 'm' or 'n').

Maybe my | notation wasn't a good idea. Lemme try () for the undo/redo:
```
start: 1a2b3c4d5e6()
 undo: 1a2b3c4d5(e6)
 undo: 1a2b3c4(d5)e6 
 undo: 1a2b3(c4)d5e6
 redo: 1a2b3(c)4d5e6  # no move happens this time
 redo: 1a2b3c(4d)5e6
 redo: 1a2b3c4d(5e)6  # this has a different move (e) associated (rather than d, above)
 redo: 1a2b3c4d5e(6)
```
(Sorry if I'm not making sense, thanks for bearing with me).

aaxu:
Oh yeah, I understand what you're saying, but that happens since 'actions' are currently a series of changes. I was thinking that if we were to make all these into compound changes, then the optimizer wouldn't work in the case of (4b)c, but would work for many mouse movements like:
(4b)cdefg => (4b)h. However, this would make it a lot easier to redo/undo since we know exactly that 'h' happened and then (4b) happened. We would never be in a spot where we redo/undo only 4 or only b.

dschuyler:
That sounds like it would work.

aaxu:
Okay I'll try to work on changing all changes to become compound ones then.

aaxu:
Closing this PR to make a new one.

Issue #58, opened 2017-08-26
Scrolling up while cursor is on the message line or status line


aaxu:
This can be reproduced by the following steps:
1. Start a new document
2. Hold enter until you have enough lines to make the text buffer scroll up
3. Scroll up and place the cursor back on the first line.
4. Put your mouse over either the message line or status line
5. Scroll up using the mouse wheel

aaxu:
Also a question. When I scroll the mouse wheel down (towards me), I expect the text buffer to scroll upward so I can see what is below my current view. Thus, I expect mouseWheelDown and scrollUp to be called. Is this not the case for you? The naming always confuses me LOL and I just want to make sure what's happening.

aaxu:
A few more questions:
* When we scroll up/down, do we also want it to move the cursor or should the cursor/all selected text stay the same?
* What is captive cursor?
Sorry I just want to clarify all these things before changing them to make sure the design stays the same. This bug caused an error in my bookmarks so I want to track it down before  finishing up and releasing the bookmarks implementation

dschuyler:
Interesting effect/bug.

# Which way is up
What is up vs. down is a long time debate in computer UI's. I've heard there were debates on it in Xerox PARC and Apple back when scrolling UIs were first done. With mouse wheels and touchpads the debate continued. IIUC, at the heart of the debate is whether the user is scrolling the cursor (focus) or whether they are scrolling the document. Now with touchpads, imo, a self evident decision has arrived because it's more intuitive for the screen to follow the user's finger (i.e. scrolling the document) when moving (rather than going the other way). Mac (and maybe others) have an OS option to reverse the scroll direction of a mouse wheel/trackpad. (Any of the above could be wrong, that's my memory of it off the top of my head).

## Maybe app.prefs.editor['scrollDirection']
Maybe we should include an option to reverse the scroll direction.

# Captive Cursor
Other console editors, such as vi, keep the cursor on the screen when scrolling. So if the window is scrolled away from the cursor, the cursor gets 'caught' by the window border and dragged along so that it stays on-screen* (what I call captive**). GUI editors don't tend to do that, they leave the cursor in-place within the document while the window scrolls.

## ci initially acted more like a console editor
[...]

PR #95, opened 2017-10-17
Allow reversing scroll direction

adamyi:
As proposed in #58 , this pr adds an option for user to reverse scroll direction. I used Apple's naming of the current direction mode in OS X - natural direction.

P.S.
Why do we want to unselect things when scrolling down/up using mouse wheel if shift is not held? For other GUI editors, I believe the behavior is to keep the original selection; for console editor like vim, the behavior is to scroll to expand/narrow the selection.

I think we can keep the selection if shift is not pressed like other gui editors and move the selection when shift is pressed like other console editors.

### DIFF
--- a/app/actions.py
+++ b/app/actions.py
@@ -1291,7 +1291,10 @@ class Actions(app.mutator.Mutator):
   def mouseWheelDown(self, shift, ctrl, alt):
     if not shift:
       self.selectionNone()
-    self.scrollUp()
+    if app.prefs.editor['naturalScrollDirection']:
+      self.scrollUp()
+    else:
+      self.scrollDown()
 
   def scrollUp(self):
     if self.view.scrollRow == 0:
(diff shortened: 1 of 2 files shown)


dschuyler:
Re: P.S. Great idea with good reasoning. Can we have both? I'd like to support a vi(m)-like mode so in that mode it would behave like vi and in the default mode it would behave like a GUI editor. What do you think?

dschuyler:
And thanks for the PR!

adamyi:
I think it would be great if we can have both. If it's just the scrolling with text selected, instead of adding that mode settings in preferences, I would prefer to support both at the same time by using whether the user is holding the shift key when scrolling to determine which behavior to use, since it would be more handy and other behaviors when shift is pressed (like using arrow keys to change the ending index of the selection) are similar to those in the visual mode in vi. However if we have several other features that we want to support two different behaviors, we can have the vi(m)-like mode settings to set them altogether. Which one do you think that we should go with?

PR #108, opened 2017-10-26
Enhancement/compound all actions

### LINKED ISSUE #88: Undo into saved undo breaks redo chain

aaxu:
This should make undoing and redoing actions more consistent. This should also resolve #88.
Right now, For undoing and redoing, I've made it so that it will undo/redo again if it first sees a trivial action (cursor move). That is, it will undo/redo until after it sees the first non-trivial action.


dschuyler:
Thanks!

I got a crash when selecting some text and pressing ctrl+f
```
error log.py 85 exception: Traceback (most recent call last):
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/background.py", line 55, in background
    program.executeCommandList(message)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/ci_program.py", line 246, in executeCommandList
    self.focusedWindow.controller.doCommand(cmd, eventInfo)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/controller.py", line 206, in doCommand
    self.controller.doCommand(ch, meta)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/controller.py", line 81, in doCommand
    cmd()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/controller.py", line 49, in changeToFind
    self.host.changeFocusTo(self.host.interactiveFind)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/window.py", line 65, in changeFocusTo
    self.parent.changeFocusTo(changeTo)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/ci_program.py", line 255, in changeFocusTo
    self.focusedWindow.focus()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/window.py", line 205, in focus
    self.controller.focus()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/editor.py", line 297, in focus
    self.textBuffer.insert(re.escape(selection))
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/actions.py", line 1191, in insert
    self.performDelete()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/actions.py", line 62, in performDelete
    self.redoAddChange(('ds', text))
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/mutator.py", line 383, in redoAddChange
    self.__compoundChange.append(change)
error log.py 85 exception: AttributeError: 'NoneType' object has no attribute 'append'
```

I don't recall what I was doing when I got this:
```
error log.py 81 exception: 'NoneType' object has no attribute 'append'
error log.py 85 exception: Traceback (most recent call last):
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/background.py", line 55, in background
    program.executeCommandList(message)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/ci_program.py", line 246, in executeCommandList
    self.focusedWindow.controller.doCommand(cmd, eventInfo)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/controller.py", line 206, in doCommand
    self.controller.doCommand(ch, meta)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/controller.py", line 81, in doCommand
    cmd()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/controller.py", line 56, in changeToGoto
    self.host.changeFocusTo(self.host.interactiveGoto)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/window.py", line 65, in changeFocusTo
    self.parent.changeFocusTo(changeTo)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/ci_program.py", line 255, in changeFocusTo
    self.focusedWindow.focus()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/window.py", line 205, in focus
    self.controller.focus()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/editor.py", line 321, in focus
    self.textBuffer.insert(str(self.host.textBuffer.penRow + 1))
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/actions.py", line 1191, in insert
    self.performDelete()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/actions.py", line 62, in performDelete
    self.redoAddChange(('ds', text))
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/mutator.py", line 383, in redoAddChange
    self.__compoundChange.append(change)
error log.py 85 exception: AttributeError: 'NoneType' object has no attribute 'append'
```

This one I was copy/pasting lines near the end of the buffer (this one might not be related to this CL)
```
error log.py 81 exception: list index out of range
error log.py 85 exception: Traceback (most recent call last):
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/background.py", line 55, in background
    program.executeCommandList(message)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/ci_program.py", line 246, in executeCommandList
    self.focusedWindow.controller.doCommand(cmd, eventInfo)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/controller.py", line 206, in doCommand
    self.controller.doCommand(ch, meta)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/controller.py", line 81, in doCommand
    cmd()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/actions.py", line 724, in editUndo
    self.undo()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/mutator.py", line 421, in undo
    self.__undoChange(change)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/mutator.py", line 490, in __undoChange
    self.lines[self.penRow] += self.lines[self.penRow + change[1]]
error log.py 85 exception: IndexError: list index out of range
```

aaxu:
Thanks for the catches! I forgot that other controllers had a text buffer as well. I now call compound add change when changing to another controller. I'm not sure if I need to check if `changeTo.textBuffer != None`  but I would guess that it wouldn't make much sense to be able to change focus to a controller that doesn't take keyboard input, so i guess it'll serve as an assert statement. Should I put in a try except clause to just log this so we know if this happens, or just put in the check, or leave it as is?

I think this should fix the first 2 errors. I cant seem to reproduce the 3rd error message. Copy and pasting shouldn't execute that line, since that seems like its undoing a carriage return.

dschuyler:
copy/past error:
in a file with these lines
```

one
two

```
1. Line select "one\ntwo\n"
2. Copy
3. Paste two times
4. Undo four times
On the fourth undo it will fail.
Also, I was able to reproduce this without this CL. So this CL is not the cause. 

I pasted the above into a new Issue #109. I'll fix it unless you're interested in it (please let me know).

aaxu:
Oh alright, you can fix this one, since I don't think I'll have much time until Tuesday.

aaxu:
I also noticed this, but really long words (40 characters long) turn into a light purplish color. What does this mean? 

image

dschuyler:
:( I was trying to identify SHA-1 strings. So if it's a mess of letters that is 40 characters (exactly), it will assume it's a SHA-1. There's also a check for 32 character sequences.

Also: I just put up changes to the --log window arrangement. That may be helpful with the undo chain. The undo chain is in the lower right and can now show more of the chain (if the window is large). I move the logging to the lower left to make room.

dschuyler:
Well actually is that a :( or a :) does recognizing common has sequences useful (it is for me, but I'd like the editor to be useful for others too).

aaxu:
Oh I mean I don't think I'll be using 40 char strings a lot so it sounds like a good feature to keep around just in case LOL. I'm all up for keeping it, though I'm not sure if we have a document to detail all our changes, so I or someone else doesn't think a feature is actually a bug.

dschuyler:
I get this when I press ctrl+o

error log.py 81 exception: 'NoneType' object has no attribute 'append'
error log.py 85 exception: Traceback (most recent call last):
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/background.py", line 55, in background
    program.executeCommandList(message)
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/ci_program.py", line 249, in executeCommandList
    self.focusedWindow.controller.onChange()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/editor.py", line 172, in onChange
    self.host.textBuffer.editPasteLines(tuple(clip))
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/actions.py", line 705, in editPasteLines
    self.performDelete()
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/actions.py", line 62, in performDelete
    self.redoAddChange(('ds', text))
error log.py 85 exception:   File "/Users/dschuyler/github/ci_edit/app/mutator.py", line 383, in redoAddChange
    self.__compoundChange.append(change)
error log.py 85 exception: AttributeError: 'NoneType' object has no attribute 'append'
error background.py 79 background: bg thread exception 'NoneType' object has no attribute 'append'

dschuyler:
ctrl+p also crashes,
but ctrl+e doesn't

aaxu:
These should be fixed now.

dschuyler:
Cool. It's been looking good so far.

Hey wdyt of starting new saved redo content for existing users?
Maybe we can change the key names in these dictionaries
```
          self.fileHistory['redoChain'] = self.redoChain
          self.fileHistory['savedAtRedoIndex'] = self.savedAtRedoIndex
```
To (for lack of something better)
```
          self.fileHistory['redoChainV2'] = self.redoChain
          self.fileHistory['savedAtRedoIndexV2'] = self.savedAtRedoIndex
```
The goal is to avoid a user having trouble by loading old-formatted redoChain from the history and trying to use it with the new redoChain.

The old 'redoChain' contents would still be there, they'd just be ignored.

aaxu:
Okay, so I changed them to redoChainCompound and savedAtRedoIndexCompound just so we remember that its from this change in implementation to compounded changes.
I also had to change __oldRedoIndex to oldRedoIndex. I'm not sure why, but it seems that when it is named __oldRedoIndex, the value written to it in `restoreUserHistory` doesn't seem to persist, so it stays at 0 and truncates the entire redo chain upon reopening the file and making a new change.

One thing I also noticed was if you have this series of moves:
* (insert)
* (new_line)
and then you undo new_line and then you type, this new change will be merged with insert, so if you undo again, you will undo the initial insert change as well. Do we want to separate these or keep them merged?

dschuyler:
> ...change __oldRedoIndex to oldRedoIndex...

Yeah it's fun little trick in the Python naming scheme. If a member starts with two underscores (without the trailing underscores, like __init__) then it's 'private' to that class. I quoted the private because it's by cooperation, not by enforcement. It's to help ya notice that __oldRedoIndex should only be used in Mutator and not in Actions. Taking off the underscores effectively moves it from a private member to a public member. Trivia: the private members are simply prefixed with the class behind the scenes. The bit-o-code below may illustrate how to access a private member in the interactive prompt.
```
>>> class Foo:
...   def __init__(self):
...     self.__bar = 5
...
>>> a=Foo()
>>> dir(a)
['_Foo__bar', '__doc__', '__init__', '__module__']
>>> a.__bar
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: Foo instance has no attribute '__bar'
>>> a._Foo__bar
5
```
Also: changing the name to remove the underscores or adding a public function to access the member are both cool with me. Accessing the private member from actions.py with 
```
self._Mutator__oldRedoIndex = self.savedAtRedoIndex  # Bad
```
would have worked, but would have been a bad choice :)

aaxu:
Oh I didn't know they weren't inherited. Good to know! Thanks for the lesson :+1: It'll probably make debugging a a bit easier now since I didn't know how to access those before LOL

dschuyler:
> Do we want to separate these or keep them merged?

The current behavior (where they are merged) seems okay. I think I'd be okay either way. Let's leave it as is and use it for awhile to see if we get a strong opinion later (unless you have a strong opinion on it now).
Trivia: The reverse order will merge new_lines
(new_line)
(insert)
undo
(new_line)
There is now only one undo item (with two new_lines)

Now for a slightly different topic: At some point I'd like to provide an redo-tree (rather than an redo-chain). The tree would hold *all* changes. The tree would not be truncated like the redo-chain is. Instead, a new branch would start. 
So the tree would look like a chain at first (which data trees do, in general - they are a list until a bifurcation occurs). After the 'undo (new_line)' above, the tree would have two branches (all the prior states of the document can be reached).
This is a future enhancement and is *not* needed for this PR.

aaxu:
Sounds good! Made a new issue for the redo tree.

dschuyler:
I've been trying this PR out and I hit something. I don't have too many details right how, so this will be vague, sorry. I edited a file. Then switched to a different git branch with that file. Then switched back to the first branch and the file appeared garbled. I disabled the 'saveUndo' and it seems okay now. This issue may be unrelated to this PR -- I should have time to check on this more later tonight or tomorrow morning. I thought I'd mention it as a heads-up, in case you had a chance to look at it.

app/mutator.py
```diff
@@ -63,17 +66,45 @@ def __init__(self):
     self.savedAtRedoIndex = 0
     self.shouldReparse = False
 
-  def compoundChangeBegin(self):
-    assert self.__compoundChange is None
+  def compoundChangeReset(self):
```
dschuyler: I'm trying to think of another name for ```compoundChangeReset```. It does 'reset', but it does much more before that. Maybe 
```compoundChangeRefresh```, or 
```compoundChangeRenew```, or 
```compoundChangeCycle```, or
```compoundChangePush```, or 
```compoundChangeAgain```.
Do any of those sound ok, or something else?

File: app/mutator.py [hunk 2]
```diff
@@ -35,7 +35,10 @@ class Mutator(app.selectable.Selectable):
   """Track and enact changes to a body of text."""
   def __init__(self):
     app.selectable.Selectable.__init__(self)
-    self.__compoundChange = None
+    self.__compoundChange = []
+    # |oldRedoIndex| is used to store the redo index before an action occurs,
+    # so we know where to insert the compound change.
+    self.oldRedoIndex = 0
```
dschuyler: how are oldRedoIndex and tempChange related (are they). They seem related, but I haven't pinned down there are two things that seem to be doing a similar thing (if they are similar).

aaxu: They are kind of related. oldRedoIndex stores the current position of the redoIndex before the new action is handled. This is because if an action contains 5 changes, we redo all 5 of them, then we go back and combine them into one and insert it at oldRedoIndex. However, if there was a tempChange, we have to insert it as its own separate change and add it to the redoChain and update oldRedoIndex by adding 1.

app/mutator.py
aaxu: I like compoundChangePush. Unless you want more descriptive names that can be longish like
compoundChangePushAndReset or something along those lines.

app/mutator.py
```diff
@@ -63,17 +66,45 @@ def __init__(self):
...
+      handledChange = False
+      # Combine changes. Assumes d, i, n, and m consist of only 1 change.
+      if (len(self.redoChain) and self.redoChain[-1][0][0] == change[0] and
+          len(self.redoChain[-1]) == 1):
+        if change[0] in ('d', 'i'):
+          change = (change[0], self.redoChain[-1][0][1] + change[1])
+          self.redoChain[-1] = (change,)
+          handledChange = True
+        elif change[0] == 'n':
+          newMouseChange = change[2]
```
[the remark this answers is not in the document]
aaxu: Oh yeah woops, I keep getting confused because people always say cursor is what your mouse moves. Will update everything in a few hours!

PR #121, opened 2017-12-06
Check if file is writable in order to send a more helpful error message

aaxu:
I don't know if `color=3` is supposed to make it red, but it doesn't look like that feature is working right now.

### DIFF
--- a/app/actions.py
+++ b/app/actions.py
@@ -952,7 +952,7 @@ class Actions(app.mutator.Mutator):
         self.setMessage('File saved')
       except Exception as e:
         self.isReadOnly = not os.access(self.fullPath, os.W_OK)
-        color = app.color.get(3)
+        color = app.color.get('status_line_error')
         if self.isReadOnly:
           self.setMessage("Permission error. Try modifing in sudo mode.",
                           color=color)
--- a/app/actions.py
+++ b/app/actions.py
@@ -951,13 +951,15 @@ class Actions(app.mutator.Mutator):
         self.fileStat = os.stat(self.fullPath)
         self.setMessage('File saved')
       except Exception as e:
+        self.isReadOnly = not os.access(self.fullPath, os.W_OK)
+        color = app.color.get(3)
         if self.isReadOnly:
           self.setMessage("Permission error. Try modifing in sudo mode.",
-                          color=3)
+                          color=color)
         else:
           self.setMessage(
               'Error writing file. The file did not save properly.',
-              color=3)
+              color=color)
         app.log.error('error writing file')
         app.log.exception(e)
     except Exception:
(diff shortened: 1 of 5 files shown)


app/mutator.py
```diff
@@ -139,7 +139,6 @@ def isDirty(self):
   def isSafeToWrite(self):
     if not os.path.exists(self.fullPath):
       return True
-    self.isReadOnly = not os.access(self.fullPath, os.W_OK)
```
dschuyler: Try making a file read only; open and edit it; try to save and get error; then change it to writable (in another terminal window); then come back to ci_edit; I'd like it to no longer show the file as read only.

(and I think this PR will still say [RO] because we only set it in fileLoad)

app/actions.py
```diff
@@ -953,9 +951,13 @@ def fileWrite(self):
...
-        self.setMessage(
-            'Error writing file. The file did not save properly.',
-            color=3)
+        if self.isReadOnly:
+          self.setMessage("Permission error. Try modifing in sudo mode.",
+                          color=3)
+        else:
+          self.setMessage(
+              'Error writing file. The file did not save properly.',
+              color=3)
```
dschuyler: Sorry, I think the color parameter is ignored right now. That could be fixed by using it in ci_edit/app/window.h around
```
      statusLine = tb.message[0]
```
By doing something like
```
      statusLine = tb.message[0]
      color = tb.message[1]
```

app/actions.py
```diff
@@ -953,9 +951,13 @@ def fileWrite(self):
         self.fileStat = os.stat(self.fullPath)
         self.setMessage('File saved')
       except Exception as e:
-        self.setMessage(
-            'Error writing file. The file did not save properly.',
-            color=3)
+        if self.isReadOnly:
+          self.setMessage("Permission error. Try modifing in sudo mode.",
+                          color=3)
```
dschuyler: Try changing these to use app.color, like
```
          self.setMessage("Permission error. Try modifing in sudo mode.",
                          color=app.color.get('error'))
```

app/mutator.py
aaxu: I think in this case we might need a background thread that just constantly polls the file for its permissions, and if the file doesnt exist, we would have to poll the directory that it is in. I think currently the [RO] will only change when you try to save, but since I don't think we can detect whether the user switched focus from some other window to ci_edit without a keylogger, I think we'd just have to check periodically.

app/actions.py
aaxu: Fixed colors for status line. Is the status line the only window that uses `setMessage()`? If not, I may need to override the function, since I'm giving it a default display color. Otherwise, every call to setMessage would need to specify the color and there will be no default. It will be more verbose, but maybe better for clarity?

app/actions.py
aaxu: Done

Issue #122, opened 2017-12-06
Select current line feature is broken


aaxu:
It starts over on the next line if you use it consecutively. Default command to do this is CTRL_L

aaxu:
Looking into this, unless you're already working on a fix.

Issue #109, opened 2017-10-27
copy/past error


dschuyler:
in a file with these lines
```

one
two

```
1. Line select "one\ntwo\n"
2. Copy
3. Paste two times
4. Undo four times
On the fourth undo the editor will exit.

aaxu:
Clarification: selecting "one\n\two\n" means you select the first two lines (not the empty third line). If you line select lines 0-2 and copy and paste that, you'll get one extra line.

PR #124, opened 2017-12-08
Bugfix/cursor select line

aaxu:
This should fix cursorSelectLine, clicking line numbers, and not break #109 .

### DIFF
--- a/app/actions.py
+++ b/app/actions.py
@@ -1412,10 +1412,13 @@ class Actions(app.mutator.Mutator):
 
     Args:
       row (int): the specified line of text that you want to select.
-    
+
     Returns:
       None
     """
+    if row >= len(self.lines):
+      self.selectionNone()
+      return
     if row + 1 < len(self.lines):
       self.cursorMoveAndMark((row + 1) - self.penRow, -self.penCol,
           0, -self.markerCol,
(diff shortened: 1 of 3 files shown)


aaxu:
Should close #122

aaxu:
I don't know if it's just me, but shift clicking doesn't seem to be working for anything, even if I'm on the master branch.

dschuyler:
Shift clicking:

It's not working for me either.

aaxu:
I've gone back several months back worth of commits and it didn't seem to be working either. Do we know if it's detectable by the program? It looks like it might be intercepted by the terminal as a way to "select" over any program.

dschuyler:
There's a behavior I was trying to mimc in the line selection. Imagine clicking on line number for line 2 in the example and then dragging into the document area onto the 'm'.
```
  1 cat
  2 dog
  3 hamster
```
The selection should stop at the 'm' rather than going to the 'r'. The PR appears to have the extension go to the 'r'.

(I'm also up for discussing whether this is good or not. I think it was the thing I was fixing when I broke the ctrl+L line selection).

dschuyler:
On shift clicking: it works in Mac OSX, but didn't work for me on Ubuntu.

aaxu:
I actually don't see it going to the 'r'. For me, when I drag my mouse to the 'm', nothing happens and the selection does not change unless my mouse goes to the line after (so the invisible line 4 in this case).

PR #125, opened 2017-12-08
Enhancement/bookmark indicator

aaxu:
I've made it so that every time you add a new bookmark, it will have a different color. I wanted to have the indicator to the left of the line number, but i would need to figure out how that would work and what to do if the line number is > 5 digits large. However, here is a beta implementation. Let me know what you think.

### DIFF
--- a/app/actions.py
+++ b/app/actions.py
@@ -98,16 +98,16 @@ class Actions(app.mutator.Mutator):
     Returns:
       A color (int) for a new bookmark.
     """
-    badColors = [103, 106, 107, 111, 121, 122]
+    badColorIndices = [103, 106, 107, 111, 121, 122]
+    nextIndex = self.nextBookmarkColorIndex
     while 1:
-      if self.nextBookmarkColor >= app.color.get('bookmarkIndexStart') + 32:
-        self.nextBookmarkColor = app.color.get('bookmarkIndexStart')
-      if self.nextBookmarkColor in badColors:
-        self.nextBookmarkColor += 1
+      if nextIndex >= app.prefs.color['bookmarkColorIndexStart'] + 32:
+        nextIndex = app.prefs.color['bookmarkColorIndexStart']
+      if nextIndex in badColorIndices:
+        nextIndex += 1
       else:
-        bookmarkColor = self.nextBookmarkColor
-        self.nextBookmarkColor += 1
-        return bookmarkColor
+        self.nextBookmarkColorIndex = nextIndex + 1
+        return nextIndex
 
   def dataToBookmark(self):
     """
@@ -129,7 +129,7 @@ class Actions(app.mutator.Mutator):
       'marker': (self.markerRow, self.markerCol),
       'pen': (self.penRow, self.penCol),
       'selectionMode': self.selectionMode,
-      'color': self.getBookmarkColor()
+      'colorIndex': self.getBookmarkColor()
     }
     upperRow, _, lowerRow, _ = self.startAndEnd()
     bookmarkRange = (upperRow, lowerRow)
--- a/app/actions.py
+++ b/app/actions.py
@@ -99,20 +99,12 @@ class Actions(app.mutator.Mutator):
       A color (int) for a new bookmark.
     """
     if app.prefs.startup['numColors'] == 8:
-      badColorIndices = [0, 7]
-      colorSetSize = 8
+      goodColorIndices = [1, 2, 3, 4, 5]
     else:
-      badColorIndices = [103, 106, 107, 111, 121, 122]
-      colorSetSize = 32
-    nextIndex = self.nextBookmarkColorIndex
-    while 1:
-      if nextIndex >= app.prefs.color['bookmarkColorIndexStart'] + colorSetSize:
-        nextIndex = app.prefs.color['bookmarkColorIndexStart']
-      if nextIndex in badColorIndices:
-        nextIndex += 1
-      else:
-        self.nextBookmarkColorIndex = nextIndex + 1
-        return nextIndex
+      goodColorIndices = [97, 98, 113, 117, 127]
+    self.nextBookmarkColorPos = (
+        self.nextBookmarkColorPos + 1) % len(goodColorIndices)
+    return goodColorIndices[self.nextBookmarkColorPos]
 
   def dataToBookmark(self):
     """
--- a/app/actions.py
+++ b/app/actions.py
@@ -87,9 +87,33 @@ class Actions(app.mutator.Mutator):
       self.getText(upperRow, upperCol, lowerRow, lowerCol)))
     self.redo()
 
+  def getBookmarkColor(self):
+    """
+    Returns a new color by cycling through a predefined
+    section of the color palette.
+
+    Args:
+      None.
+
+    Returns:
+      A color (int) for a new bookmark.
+    """
+    badColors = [103, 106, 107, 111, 121, 122]
+    while 1:
+      if self.nextBookmarkColor >= app.color.get('bookmarkIndexStart') + 32:
+        self.nextBookmarkColor = app.color.get('bookmarkIndexStart')
+      if self.nextBookmarkColor in badColors:
+        self.nextBookmarkColor += 1
+      else:
+        bookmarkColor = self.nextBookmarkColor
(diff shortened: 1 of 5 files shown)


dschuyler:
Try making several bookmarks in a large-ish file (more than one screen) that will cause the content to scroll while visiting bookmarks. The bookmark highlighting doesn't seem to move with the bookmark.

app/actions.py
```diff
@@ -84,9 +84,33 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
       self.getText(upperRow, upperCol, lowerRow, lowerCol)))
     self.redo()
 
+  def getBookmarkColor(self):
```
dschuyler: Does this work well in an 8 color terminal? Heh, well, maybe not work well, but does it not crash and not render super badly (other than how bad things look with only 8 colors)?

app/default_prefs.py
```diff
@@ -153,6 +153,7 @@ def numberTest(str, expectRegs):
 
 color256 = {
   '_pre_selection': stringColorIndex,
+  'bookmarkColorIndexStart': 96,
```
dschuyler: I added a check (assert) to look for an color256 and color8 getting out of sync. (so this should assert on startup unless this is added to color8 as well).

aaxu: Yep, 8-bit is supported now!

aaxu:
Note: I had to move the call to **setUpPalette** to after **parseArgs** was called, since I needed to check for the --eightBit flag. It doesn't seem to cause any issues as far as I'm aware though.

dschuyler:
Tip: it's 8 colors rather than 8-bit.

Trivia: 8-bit color has 256 colors. 3-bit color would have 8 colors. In video games the term 8-bit color is used, but in the terminal '256 color' is used (even though it means the same thing). So we don't say 3-bit color for a similar reason, we say 8 color (even though it's 3-bit color).

aaxu:
Ohh okay, changed! Thanks.

app/color.py
```diff
@@ -30,12 +30,8 @@ def get(colorType, delta=0):
...
   colorIndex = min(colors - 1, colorIndex + delta)
-  #colorIndex = colorIndex % colors
-  r = cache__.get(colorIndex)
-  if r is not None:
-    return r
-  color = curses.color_pair(colorIndex)
+  color = cache__.get(colorIndex) or curses.color_pair(colorIndex)
+  cache__[colorIndex] = color
   if colorType in ('error', 'misspelling'):
     color |= curses.A_BOLD | curses.A_REVERSE
```
dschuyler: What is better about the new code?
At first glance it looks like it will repeatedly reset the same cache value; and repeatedly check for 'error' or 'misspelling'.

aaxu:
Before, you would check if an index is in the cache, if it is, return its cached value. However, if we were to load a misspelling first, (curses.color_pair(3) then modify by ORing some bits), we will cache this modified value. Thus all future references to index color 3 will be the modified value, not the set base value. In the new changes, we will repeatedly set the cache value, but it will be setting the cache value to itself once it has been set. It functions the same as `dict.setdefault(key, defaultVal)` and prevents the need to use if statements.
This way, we always only store the base color value, and then we can add any modifications afterward.

app/color.py
aaxu: Accidentally put the answer to this question below.

aaxu:
Has there been any updates on this PR, @dschuyler? I think the changes have been fixed.

app/actions.py
```diff
@@ -86,9 +86,38 @@ def performDeleteRange(self, upperRow, upperCol, lowerRow, lowerCol):
...
+      colorSetSize = 32
+    nextIndex = self.nextBookmarkColorIndex
+    while 1:
+      if nextIndex >= app.prefs.color['bookmarkColorIndexStart'] + colorSetSize:
+        nextIndex = app.prefs.color['bookmarkColorIndexStart']
+      if nextIndex in badColorIndices:
+        nextIndex += 1
+      else:
+        self.nextBookmarkColorIndex = nextIndex + 1
+        return nextIndex
```
dschuyler: wdyt of using a goodColorIndices list?
```
goodColorIndices = [104, 105, 108, 109, 110]
if app.prefs.startup['numColors'] == 8:
  goodColorIndices = [1, 2, 3, 4, 5]
self.nextBookmarkColorIndex = (
    self.nextBookmarkColorIndex + 1) % len(goodColorIndices)
return goodColorIndices[self.nextBookmarkColorIndex]
```

app/ci_program.py
```diff
@@ -592,6 +592,8 @@ def parseArgs(self):
         elif i == '--clearHistory':
           app.history.clearUserHistory()
           self.quitNow()
+        elif i == '--eightColors':
+          numColors = 8
```
dschuyler: Thanks for adding the --eightColors switch. Consider making this its own PR and we can land it straight away (and it would make this PR smaller, which is nice).

app/default_prefs.py
dschuyler: Thanks!

app/window.py
```diff
@@ -362,17 +363,53 @@ def __init__(self, host):
...
+  def bookmarkForRow(self, row):
+    """
+    Checks whether the specified row lies on a bookmark or not.
+
+    Args:
+      row (int): the row number you want to check between [0, numLines - 1].
+
+    Returns:
+      The bookmark that the row is associated with or None if there isn't one.
+    """
```
dschuyler: Since bookmarks tend to come in blocks of rows, would it be more efficient to have a function that returned a list of bookmarked rows on the current screen?

i.e. 
- args are beginRow, endRow
- use bisect to find beginRow
- then 'walk' the bookmarks until endRow
- return list of bookmarked rows

app/ci_program.py
aaxu: Okay I've created a new PR, though I'm not really sure of a clean way to remove all relevant changes without manually doing it and having manual cleanup.

dschuyler: I've just merged the other PR. I *think* doing a merge to head will remove the changes shown in this PR (that are in that PR).

app/actions.py
aaxu: I guess this would work well, and we can limit it to ~5ish colors each so it would be easier to maintain (so we don't have to rewrite up to 32 colors if we decide on a color scheme change).

app/window.py
aaxu: I ended up returning a list of bookmarks itself, since we need the actual bookmarks when rendering the line numbers. I think it should perform a bit better now.

app/ci_program.py
aaxu: Nice!

app/color.py
dschuyler: Thanks for explaining. I'm going to go ahead with this PR, but I'll likely want to change this code again. Here's why: in some coding, avoiding if-statements is important for performance, in other situations, avoiding function calls is more important.

On one extreme, in assembly on a Cell processor, like in the PlayStation 3, avoiding if-statements (conditional-branching) is critical for high performance; while a calling a function (unconditional-branch) is close to free by comparison.

On another extreme, in Python (and many other scripting languages) avoiding function calls is critical for high performance; while if-statements are close to free by comparison.

In many environments, (in both extremes above) calling out to an external (non-local) library is very time consuming. One of the main reasons for color.get() to exist is to reduce the number of calls out to curses.color_pair() since calling into curses from Python is time consuming.

dschuyler:
Hey, are ya interested in writing a unit test for the bookmark indicators?

aaxu:
Yes I am, but I'm not really sure how to exactly go about it yet. I think I can test getBookmarkColor, but I'm not sure how to test if the LineNumber's render function works correctly, or even if the getVisibleBookmarks function works correctly, since it seems to reference the InputWindow's textBuffer's bookmarks. This seems like I could only do it through an integration test and I don't have any experience doing that in Python.

aaxu:
If I can end up figuring out how the whole program works and how to test everything, I wouldn't mind writing a lot of the testing in general. Any tips would be appreciated! I'm looking at some of the already written ones which use a fake curses window and fake view, but I'll still need to play around with it before I can get something working.

PR #130, opened 2017-12-22
Enhancement/check external file changes

aaxu:
New changes:
* textBuffer objects now have a FileStats object which is used to get information about the file that it refers to. Assuming that `renameBuffer()` is used to change the fullPath of a text buffer, the FileStats object will also be changed to track the new file.
* The FileStats object polls the file on disk every 2 seconds to see if there are any changes. If it detects that the write bit has been changed, it will refresh the screen so that [RO] gets appended/removed from the top of the screen. If the program is running in single-threaded mode, the object will not automatically update, but you can update it manually, or call `getUpdatedFileInfo` which returns updated information about the file.

I plan on expanding this to detect if the file has been changed on disk since we have opened it, but I wanted to check in on what you think about this implementation first.


app/window.py
```diff
@@ -578,7 +578,7 @@ def onChange(self):
           lineCursor -= 1
     pathLine = self.host.textBuffer.fullPath
     if 1:
-      if tb.isReadOnly:
+      if tb.fileStats.getUpdatedFileInfo()['isReadOnly']:
```
dschuyler: This can be called frequently (when scrolling the window for example). Acquiring a semaphore and doing some file IO can be heavy. What happens if we use the saved (cached) value? It could be slightly out of date, but that won't last long. I think I'd prefer to have some lag on updating the RO state (happens infrequently) than lag when scrolling (happens frequently).

app/background.py
```diff
@@ -46,22 +46,39 @@ def put(self, data):
...
+    Sends a SIGUSR1 signal to the current program and draws its screen.
+
+    Args:
+      program (CiProgram): an instance of the CiProgram object.
+
+    Returns:
+      None.
+    """
+    pid = os.getpid()
+    signalNumber = signal.SIGUSR1
```
dschuyler: How about moving these above this def (so they are only done once, rather than with each call to redrawProgram)
```
   pid = os.getpid()
   signalNumber = signal.SIGUSR1
```

app/background.py
```diff
@@ -46,22 +46,39 @@ def put(self, data):
...
+    os.kill(pid, signalNumber)
+
   block = True
-  pid = os.getpid()
-  signalNumber = signal.SIGUSR1
   while True:
     try:
       try:
-        program, message = inputQueue.get(block)
+        program, message, callerSemaphore = inputQueue.get(block)
```
dschuyler: Does passing an empty list for message also cause a redraw?

aaxu: Ahh it does. Then I guess we could remove the 'redraw' message.

app/window.py
aaxu: Oh that sounds good. I did this because it also supported single threading nicely. Should be fixed now.

app/background.py
aaxu: Sounds good!

aaxu:
I actually have the working extension of this PR which allows the program to detect if the file on disk has changed (for example if you change branches), and creates a popup box to ask if you want to reload the file. It refactored and changed the implementation of some of the code included in this PR, but it does include a lot of changes, so I don't know if you want to review this PR first, then the extension, or if I should replace this PR with the extension entirely.

dschuyler:
How about a new PR with the full changes (which would be a replacement). It would be great to see the full change PR before deciding if it would be better done in pieces.

Though in general, a smaller PR is better most of the time. I know that's tricky to do sometimes.

aaxu:
Okay sounds good. I just wanted to try and see if the extension of this PR would work since that is the main purpose of the background thread. I'll put something up tonight once i move over some commits and clean it up a bit.

PR #140, opened 2018-01-17
Tests and bug fix for bookmark indicators

aaxu:
This adds tests for the `getBookmarkColor` function and the `getVisibleBookmarks` function. Still have yet to make tests for render, but I just wanted to check if these looked good and if there were anhy things I should do to make the tests better.

Also, since we are using python 2.7, I had to install the **mock** library manually, since I used it for my tests.

### DIFF
--- a/app/unit_test_bookmarks.py
+++ b/app/unit_test_bookmarks.py
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc.
+# Copyright 2018 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -11,18 +11,34 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-import mock
+import os
+import sys
 import unittest
 
+from app.curses_util import *
+import app.fake_curses_testing
 import app.prefs
 import app.text_buffer
 import app.window
 
+kTestFile = '#bookmarks_test_file_with_unlikely_file_name~'
 
-class BookmarkTestCases(unittest.TestCase):
+class EmptyObject:
+  pass
+
+class BookmarkTestCases(app.fake_curses_testing.FakeCursesTestCase):
   def setUp(self):
-    self.fakeHost = mock.Mock()
+    if True:
+      # The buffer manager will retain the test file in RAM. Reset it.
+      try:
+        del sys.modules['app.buffer_manager']
+        import app.buffer_manager
+      except KeyError:
+        pass
+    if os.path.isfile(kTestFile):
+      os.unlink(kTestFile)
+    self.assertFalse(os.path.isfile(kTestFile))
+    self.fakeHost = EmptyObject()
     self.textBuffer = app.text_buffer.TextBuffer()
     self.textBuffer.lines = 50
     self.lineNumbers = app.window.LineNumbers(self.fakeHost)
@@ -31,11 +47,30 @@ class BookmarkTestCases(unittest.TestCase):
     self.fakeHost.lineNumberColumn = self.lineNumbers
     self.fakeHost.textBuffer = self.textBuffer
     self.fakeHost.scrollRow = self.fakeHost.cursorRow = 0
+    app.fake_curses_testing.FakeCursesTestCase.setUp(self)
 
   def tearDown(self):
-    pass
+    app.fake_curses_testing.FakeCursesTestCase.tearDown(self)
+
+  def runWithTestFile(self, fakeInputs):
+    sys.argv = [kTestFile]
+    self.assertFalse(os.path.isfile(kTestFile))
+    self.runWithFakeInputs(fakeInputs)
 
   def test_get_next_bookmark_color(self):
+    try:
+      import mock
+    except ImportError:
+      startYellow = '\033[93m'
+      disableColor = '\033[0m'
+      startBlue = '\033[94m'
+      exceptionMessage = (startYellow + "This test could " +
+          "not execute because the 'mock' module could not be found. If " +
+          "you would like to run this test, please install the mock module " +
+          "for python 2.7. You can visit their website at " + startBlue +
+          "https://pypi.python.org/pypi/mock " + startYellow + "or you can " +
+          "try running " + startBlue + "pip install mock." + disableColor)
+      raise Exception(exceptionMessage)
     def test_with_an_x_colored_terminal(x):
       mock.patch.dict(app.prefs.startup, {'numColors': x}, clear=True)
       colors = set()
@@ -59,14 +94,16 @@ class BookmarkTestCases(unittest.TestCase):
       test_with_an_x_colored_terminal(256)
 
   def test_get_visible_bookmarks(self):
-    # Set up mock objects to test the LineNumbers methods.
-
+    # Set up the fake objects to test the LineNumbers methods.
     self.textBuffer.bookmarks = [((0, 0),), ((10, 10),), ((20, 20),),
                                  ((30, 30),), ((40, 40),),]
     visibleBookmarks = self.lineNumbers.getVisibleBookmarks(
         self.fakeHost.scrollRow, self.lineNumbers.rows)
     expectedBookmarks = {((0, 0),), ((10, 10),), ((20, 20),)}
+
+    # Check that visibleBookmarks contains all the correct bookmarks
     self.assertEqual(set(visibleBookmarks), expectedBookmarks)
+    # Check that the number of bookmarks is the same, as set removes duplicates.
     self.assertEqual(len(visibleBookmarks), len(expectedBookmarks))
 
     self.fakeHost.scrollRow = 20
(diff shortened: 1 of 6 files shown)


app/unit_test_bookmarks.py
```diff
@@ -0,0 +1,106 @@
+# Copyright 2016 Google Inc.
```
dschuyler: 2018

app/unit_test_bookmarks.py
```diff
@@ -0,0 +1,106 @@
...
+  def test_get_visible_bookmarks(self):
+    # Set up mock objects to test the LineNumbers methods.
+
+    self.textBuffer.bookmarks = [((0, 0),), ((10, 10),), ((20, 20),),
+                                 ((30, 30),), ((40, 40),),]
+    visibleBookmarks = self.lineNumbers.getVisibleBookmarks(
+        self.fakeHost.scrollRow, self.lineNumbers.rows)
+    expectedBookmarks = {((0, 0),), ((10, 10),), ((20, 20),)}
+    self.assertEqual(set(visibleBookmarks), expectedBookmarks)
+    self.assertEqual(len(visibleBookmarks), len(expectedBookmarks))
```
dschuyler: Can line 67 succeed where line 68 fails? I.e. can the lengths be unequal where the sets are equal?

app/unit_test_bookmarks.py
```diff
@@ -0,0 +1,106 @@
...
+import app.window
+import mock
+import unittest
+
+class BookmarkTestCases(unittest.TestCase):
+  def setUp(self):
+    self.fakeHost = mock.Mock()
+    self.textBuffer = app.text_buffer.TextBuffer()
+    self.textBuffer.lines = 50
+    self.lineNumbers = app.window.LineNumbers(self.fakeHost)
```
dschuyler: While mock is handy, it looks like this use can be replaced with a stub class. I'd like to avoid adding a dependency if it can be avoided.

Where there are dependencies, they are put in the third_party directory. That was done for pyperclip since pasteboard support is fairly fundamental (I haven't found a way around it). WDYT about not using mock here?

app/unit_test_bookmarks.py
aaxu: Although this is a side case, this would happen if visibleBookmarks contains duplicates.

app/unit_test_bookmarks.py
aaxu: Yes, I agree that these could be using a stub class instead. I just used them because I ended up having to import the mock library anyway, so it would look nicer to use a mock. One of the issues I was trying to solve was that `getBookmarkColor` references **app.prefs.startup**. For my tests, I wanted to set this value in order to test for the 8 and 256 colored terminals. However, I was thinking that if I were to change the dictionary directly and the tests were run in parallel (not really sure how they are run in github, or maybe elsewhere in the future), then this change could affect some other tests.

Maybe I was overthinking it and so I used a mock, but I also thought that since its for only used for running tests and it wouldn't affect the performance of the program, it would be okay. I could switch these mocks to be stub classes anyway in order to make the individual tests run faster, but I still think I need the `mock` library as I haven't found another way to stub out a dictionary like app.prefs.startup, besides mock.patch.

app/unit_test_bookmarks.py
aaxu: Oh haha okay

app/unit_test_bookmarks.py
dschuyler: Oh, because set() will remove duplicates. Got it.
Maybe add a comment like
# Check length because set() will remove duplicates.
or something like that.

aaxu: Sure, sorry for the confusion!

app/unit_test_bookmarks.py
aaxu: Okay, I can try that. I thought that this wouldn't cause any trouble, since we could add this library in setup.py or something, which would end up minimizing any interaction the user has with installing any external dependencies. Perhaps even a docker image for developers to make it even easier. I'll try this exclusion method for now.

aaxu: Updated.

dschuyler:
Hi, I've just checked in ci_edit/app/fake_curses_testing.py which has a base class to help with checking the display text and the color/style. It still needs documentation, but there's examples using it in ci_edit/app/unit_tests_application.py.

aaxu:
Oh great! This opens up a whole new set of doors for me in testing bookmarks. However, I have a few questions about this new module.

* It looks like the `displayCheck` requires the following line of code in setUp or else it doesn't operate properly:
```
if True:
      # The buffer manager will retain the test file in RAM. Reset it.
      try:
        del sys.modules['app.buffer_manager']
        import app.buffer_manager
      except KeyError:
        pass
``` 
Can you tell me a bit more about why that is? Should we refactor it and just put this in `fake_curses_testing.py.setUp`?
* When I fail a `displayCheck` call, I get an error message which resembles this.
image
I'm guessing that the right hand side is supposed to be the actual output and the left hand side is the expected output? However, the right hand side seems to always stay like this. Is there a way to interpret this?

dschuyler:
About resetting buffer manager:
Modules are not reloaded by the unittest. So globals carry over from one test to the next.

About the display output:
Yeah, sorry, that's not described well. The left side is the text from the display without styling. The right side has a different character for each styling/color. So all the ^ are the same color (and style, like bold) and different from [, ], and `. The actual values are not as useful and don't show well. I'm open to ideas on how to improve this, but for now the two views are different 'layers' of the same screen.

aaxu:
does displayCheck always expect a display passed in with 11 textbuffer lines (15 total, with logo, status line,etc)? My tests are going to be so long haha

dschuyler:
>does displayCheck always expect a display passed in with 11 textbuffer lines (15 total, with logo, status line,etc)?

Nope :) Pass the start row, column in the first two parameters and only provide as much text as you want to check.

e.g. 
```
self.displayCheck(2, 7, ["text "])
```

aaxu:
Oh, okay good. I was also wondering if there was a way to change the size of the window from a test. Perhaps we could also add an enhancement where we can type out strings instead of characters and the function would split it up for us. For example, we can type `"one"` instead of `'o', 'n', 'e'`. I know python 3 can unpack tuples into arguments, but I don't think python 2 can do this :(

Also, I wrote up some tests to check if jumping to bookmarks was working properly. However, right now it's failing and I think it has to do with something in the fake display. The displayed "actual" output is wrong, but it works perfectly fine if I mimic the commands in the regular program. If you try to run these tests, you can probably see what I'm talking about.

dschuyler:
> For example, we can type "one" instead of 'o', 'n', 'e'

Okay, I just added two functions to help with this. 

- `self.setClipboard('fun')`
    - This will copy the text to the clipboard, so to get it in the document, follow-up with a `CTRL_V`
- `self.writeText('hello')`
    - This will call setClipboard and do the `CTRL_V` for you (as a convenience).

aaxu:
Nice! Do you have any idea about why the fake display is showing incorrect output? I performed exactly the keys that were in the test and the fake display seems to differ from the real one.

dschuyler:
> Do you have any idea about why the fake display is showing incorrect output?

I think that F2 is not setting the waiting for input flag. I'll look into it.

app/unit_test_bookmarks.py
```diff
@@ -104,3 +143,233 @@ def test_get_visible_bookmarks(self):
...
+            "    15 fifteen                          ",
+            "    16 sixteen                          ",
+            "                        8, 6 |  30%,100%",
+            "                                        "]),
+        KEY_F2, # Jump to the fourth bookmark (23, 12).
+        # This moves the bookmark to the optimal scroll position.
+        self.displayCheck(0, 0, [
+            " ci     .                               ",
+            "                                        ",
+            "    21  twentyone                       ",
```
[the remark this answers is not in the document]
aaxu: LOL I know but it was like saving 5 extra keystrokes.. Okay fine I'll change it haha

aaxu:
Also, I think it's kind of hard to know exactly where the display is wrong with an error message like:
`unit_test_bookmarks.py:210:test_bookmarks_jump(): row 13, col 27 mismatch '3' != '2' at index 187`
Maybe in addition to this, we could also print out both of the offending rows?

dschuyler:
> Maybe in addition to this, we could also print out both of the offending rows?

Yeah. I just checked in something that does this (how does it look):

```
AssertionError:
  unit_test_application.py:93:test_text_contents():  at index 5
  row 2, col 7 mismatch 't' != 'a'
  actual:   |     1 text                             |
  expected: |     1 aext                             |
  mismatch:         ^
```

dschuyler:
> > Do you have any idea about why the fake display is showing incorrect output?

> I think that F2 is not setting the waiting for input flag. I'll look into it.

Okay, that guess seems wrong (the code looks correct).

My new guess is that optimalCursorRow may be set differently in your prefs (vs. the unit tests).

dschuyler:
> I was also wondering if there was a way to change the size of the window from a test. 

I've just added `resizeScreen(rows, cols)` to the `FakeCursesTestCase` class. There's an example using it in `ci_edit/app/unit_test_application.py`.

```
self.resizeScreen(10, 36),
```

aaxu:
>Yeah. I just checked in something that does this (how does it look):
```
AssertionError:
  unit_test_application.py:93:test_text_contents():  at index 5
  row 2, col 7 mismatch 't' != 'a'
  actual:   |     1 text                             |
  expected: |     1 aext                             |
  mismatch:         ^
```

ooo that looks great. I also had another idea where maybe it would be similar to 
```
   display                                    style
  +----------------------------------------+ +----------------------------------------+
 0| ci     *                               | |________________________________________|
 1|                                        | |________________________________________|
 2|     1 one                              | |eeeeeee[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
 3|     2 two                              | |ggggggg[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
 4|     3 three                            | |ijjjjji[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
 5|     4 four                             | |_______[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
 6|     5 five                             | |_______[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
 7|     6 six                              | |_______[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
 8|     7 seven                            | |_______[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
 9|     8 eight                            | |_______[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
10|     9 nine                             | |kkkkkkk[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
11|    10 ten                              | |_______[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
12|    11 eleven                           | |_______[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[]|
13|                        3, 6 |   8%,100%| |________________________________________|
14|                                        | |aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa|
  +----------------------------------------+ +----------------------------------------+
FAIL

```

but instead, we would print out the actual and the expected output, and perhaps we can color both the offending lines in a different color instead? I thought this would be an even more helpful message than just printing out the two lines

aaxu:
Hey @dschuyler, sorry for being gone for a while. Been bogged down with school work. Are there any other changes you want me to make to this PR? I think the refactoring/bookmarks PR builds off this one, so I'm waiting till this gets accepted before fixing some issues in that branch.

I think the only issue right now is the testing API bug that is present in `fake_curses_testing.py` as described in some post above, but I haven't taken a look at what's causing the issue yet.

dschuyler:
Glad to see you back (got worried :) ).
I'll look into the testing code further.

dschuyler:
What are your thoughts on https://github.com/google/ci_edit/commit/278a7d7c64c6c2a6d38dfff960258ad159fdeb4c
(that's a version of this PR in a branch).

Is it correct?

I'm not confident that I can reproduce the unit test in the GUI without making an error (to verify it).

aaxu:
Yes that looks great! I've just merged with it and updated an old comment so it matches the current test.

I was wondering if this would end up causing any maintainability issues in the future. It looks like if we make any gui changes, (or specific to this test, any changes to the default value of optimalCursorValue), will require us to make changes to the `test_bookmarks_jump` test as well, and possibly any other test that depends on self.displayCheck. Is this normal, or is there a better way to do it?

It sounds hard since it looks like self.displayCheck is used for integration testing and testing the gui seems to be a large part of it. Maybe for some tests, rather than checking what actually appears on the gui, we could check the exact states of the fake curses window? For example, in this case, we could do something like:

```
self.display.getBookmarks() == [.....]
self.display.getCursorPosition() == (x, y)
self.displaly.getScrollPosition() == z * optimalCursorPosition..
```

Then it wouldn't be dependent on the gui but rather the implementation of the code.

PR #141, opened 2018-01-23
Moved popup window over from detectFileChanges branch

aaxu:
This contains the standalone code for the popup window. The DetectFileChanges branch still creates the infrastructure in the background thread and ci_program.py file to make the popup box appear through the background thread. 

This popup right now has `reloadBuffer` hardcoded to Y and `returnToHost` hardcoded to N. It also doesn't contain a text buffer that the user could type on. I plan to eventually improve this popup box so that it can branch out and also be used for things like overwriting a file and saving a new file (requires text box so that user can input a file name).


aaxu:
We talked about making the popup window a top level modal before. However, when I think about expanding this feature to saving files and such, it makes more sense for each of them to have their own instance of a popup window, so that the options do not have to be updated every time the popup shows. This would work much like how right now there are several instances of LabeledLine in the InputWindow.   What do you think about this approach vs having a global popup and manually changing its options on every popup?

dschuyler:
Having a pop up window base class sounds good.
How about removing the reload file code from this PR, so that this is just about the base class.

aaxu:
Okay, I removed the reloadBuffer function. I also reimplemented the popupWindow class a bit so that its API makes it more flexible to be used by other things. I'm thinking that maybe every window that causes a popup could have its own function. However, because the fileStats object doesn't necessarily belong directly to a window, it uses the inputWindow's own popupWindow object. As I actually don't think that there will be multiple popups firing, I think it would be okay to just have all of them reference the inputWindow's popup.

app/cu_editor.py
```diff
@@ -397,6 +397,45 @@ def setTextBuffer(self, textBuffer):
...
+    self.commandSet = {
+      KEY_ESCAPE: self.changeToMainWindow,
+    }
+
+  def changeToMainWindow(self):
+    self.view.hide()
+    self.host.changeFocusTo(self.host.host)
+    if self.callerSemaphore:
+      self.callerSemaphore.release()
+      self.callerSemaphore = None
```
dschuyler: Will enough popup instance use a callerSemaphore to make this part of the base class?

aaxu: No I would guess that only other threads would be using this semaphore. I just thought I would provide the infrastructure so that all the windows can reference the InputWindow's popup window. For the windows that do not need the semaphore, they can simply ignore it.

app/cu_editor.py
```diff
@@ -397,6 +397,45 @@ def setTextBuffer(self, textBuffer):
...
+    self.callerSemaphore = None
+    def noOp(ch, meta):
+      app.log.info('noOp in PopupController')
+    self.commandDefault = noOp
+    self.commandSet = {
+      KEY_ESCAPE: self.changeToMainWindow,
+    }
+
+  def changeToMainWindow(self):
+    self.view.hide()
```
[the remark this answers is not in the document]
aaxu: Fixed

app/cu_editor.py
```diff
@@ -396,11 +396,49 @@ def setTextBuffer(self, textBuffer):
...
+      options (dict): A dictionary mapping keys (ints) to its
+                      corresponding action.
+
+    Returns;
+      None.
+    """
+    self.commandSet = options
+
+  def setTextBuffer(self, textBuffer):
+    self.textBuffer = textBuffer
```
dschuyler: I think the base class does this. So this could be removed.

app/cu_editor.py
```diff
@@ -396,11 +396,49 @@ def setTextBuffer(self, textBuffer):
...
+    popup window as well as their functions.
+
+    Args:
+      options (dict): A dictionary mapping keys (ints) to its
+                      corresponding action.
+
+    Returns;
+      None.
+    """
+    self.commandSet = options
```
dschuyler: Maybe make a note that the caller should include esc to close (since this will replace the entry from line 416); or maybe do .update(options) rather than assignment?

app/cu_editor.py
```diff
@@ -396,11 +396,49 @@ def setTextBuffer(self, textBuffer):
       KEY_F2: textBuffer.bookmarkNext,
       KEY_F3: textBuffer.findAgain,
       KEY_F4: self.changeToPaletteWindow,
+      KEY_F5: self.changeToPopup,
```
dschuyler: A test key is okay, but consider commenting it out for check-in and/or make a note that this is for testing.

dschuyler:
I'm going to go ahead and pull this in. Feel free to follow-up on the suggestions I made above in another PR (if you like).

app/cu_editor.py
aaxu: Oops forgot to remove this. I'll edit it in the next PR. Sorry about that.

app/cu_editor.py
aaxu: Will add the note. I thought about the .update, but then the user may forget to overwrite or delete any old keys. I think it might be better to have them list it all out again when they modify it. Though, I guess we could have both an `update` and a `set` function.

PR #143, opened 2018-01-27
Made row/column percentages more consistent.

### LINKED ISSUE #142: Statistics not really accurate
Given a file with 3 lines, being on the first line will show the row % as 0%. The second line will be 33%, and the last line will be 100%. I understand how these were calculated, but I think the values should be more consistent as it increases. (Makes it easier to write tests as well)

aaxu:
Resolves #142

### DIFF
--- a/app/window.py
+++ b/app/window.py
@@ -561,7 +561,7 @@ class StatusLine(ViewWindow):
     if lineCount == 1:
       rowPercentage = 100
     else:
-      rowPercentage = self.host.cursorRow * 100 / (lineCount - 1)
+      rowPercentage = (self.host.cursorRow + 1) * 100 / lineCount
     charCount = len(tb.lines[self.host.cursorRow])
     if self.host.cursorCol == charCount:
       colPercentage = 100
--- a/app/window.py
+++ b/app/window.py
@@ -557,15 +557,16 @@ class StatusLine(ViewWindow):
     rowPercentage = 0
     colPercentage = 0
     lineCount = len(tb.lines)
-    if lineCount:
-      rowPercentage = self.host.cursorRow * 100 / lineCount
-      if self.host.cursorRow >= lineCount - 1:
-         rowPercentage = 100
-      charCount = len(tb.lines[self.host.cursorRow])
-      if (self.host.cursorCol < charCount):
-        colPercentage = self.host.cursorCol * 100 / charCount
-      else:
-        colPercentage = 100
+    # Assumes lineCount is always at least 1.
+    if lineCount == 1:
+      rowPercentage = 100
+    else:
+      rowPercentage = self.host.cursorRow * 100 / (lineCount - 1)
+    charCount = len(tb.lines[self.host.cursorRow])
+    if self.host.cursorCol == charCount:
+      colPercentage = 100
+    else:
+      colPercentage = self.host.cursorCol * 100 / charCount
     # Format.
     rightSide = ''
     if len(statusLine):


dschuyler:
The column percentages look good.
The lines seem off to me.
i.e. if there are ten characters then moving left to right goes in 10% increments.
But, going down ten lines doesn't.

### More examples

Imagine the dots are are characters in a doc, and the cursor is at the X:
```
X...
....
....
....
```
The above looks like 0%, 0% (and both versions agree, I'm just setting the stage)

```
.X..
....
....
....
```
The above looks like 0%, 25% (and both agree)

```
....
.X..
....
....
```
The above looks like 25%, 25% (but this patch is saying 33%, 25%)

```
....
....
..X.
....
```
The above looks like 50%, 50% (but this patch is saying 66%, 50%)


#### And this is where things get tricky.
At this point the old code and patch code agree, but both seem odd.
```
....
....
....
...X
```
The old code and this patch say the above is 100%, 75%, but should this be 75%, 75% instead?

```
....
....
....
....X
```
 What should this (the above) be? Again, both versions agree (100%, 100%), but is either correct?

Essentially the problem comes from it feeling strange to be on the last line of the document and not be at 100%.  For the column, there is an unseen extra position past the last character. There's no \n on the line above. The cursor is truly at the very end of the document. So jumping to 100% for the lines is a hack (in the hope that it looks nice, but may it doesn't).

### options
So IIUC, we need to choose between three imperfect (in my opinion) options:
- not showing row at 100% (unless the document is empty). This keeps the values consistent with reality, but doesn't feel satisfying to some users.
- keep the numbers consistent, except for the last line: force that to 100%. This makes a weird jump between the penultimate line and the last line. This is what the old code did. I think that's what the bug was pointing out. I agree it's weird.
- average/spread the weird jump at the end, across may lines, so the last line jump is not so large. This makes the values kinda off (but not far off) often. This PR takes that tact.

#### Questions
WDYT? 
Would the first option be better or worse than the other options in your opinion?
(I'm guessing you prefer the third option over the second one because this PR was made)
(FWIW, I kinda prefer the second option to the third. I'm trying to decide how I feel about the first option)

aaxu:
Just realized that responding via email doesn't actually create a response here. I just updated the implementation of the rows.
I made it so that basically the percentage is based on what line/column you are able to "edit".

For example, if there are 4 rows and the cursor is on the first row, then the row% would be 25. The reason for this is because you are effectively able to edit the first row (del, add, select line, etc.). Note that this also means that row% will never be 0.
The row % is still the same, though there doesn't seem to be any great reason for it other than consistency. I guess one reason would be when you press backspace, you are erasing the i'th character (hence modifying). There is the delete key as well, but I would like to imagine that the backspace key is used more (and the equivalent of backspace for macs).

This is honestly feels like I'm starting to stretch it, or it feels very arbitrary, but here's just an example on paper:


```
|    100%, 100%
``` 

```
x|    100%, 0%
``` 

```
|xxxx
xxxx
xxxx
xxxx    25%, 0%
``` 


```
x|xxx
xxxx
xxxx
xxxx    25%, 25%
``` 

```
xxxx
xx|xx
xxxx
xxxx    50%, 50%
``` 

```
xxxx
xxxx
xxx|x
xxxx    75%, 75%
``` 

```
xxxx
xxxx
xxxx
xxxx|    100%, 100%
```

dschuyler:
Would this one be 100%, 100% (rather than the 100%, 0% mentioned above)?
```
x|    100%, 100%
```

aaxu:
Ah yes my mistake. Updated my comment.

dschuyler:
What do you think of using a word like 'top' (or 'bottom') to cover the oddity? (or 'first' 'last')

So when the cursor is on the first line, the percentage would be replaced with a word like

```
|xxxx
xxxx
xxxx
xxxx    top, 0%
```

```
x|xxx
xxxx
xxxx
xxxx    top, 25%
```

```
xxxx
xxxx
xxxx
xxxx|    bottom, 100%
```

aaxu:
In that case it would go from top -> 50% -> 75% -> bottom, which still presents an oddity, though it may be one that is less noticeable. I'm good with either one for now. If there is only one line though, should we consider that the bottom? I think that would be slightly more consistent, so that if the user types in a carriage return, the display stays as 'bottom'

dschuyler:
Excellent point, but 'top' makes more sense to me. It's interesting to see where our view points differ. It's really helpful to hear a different point of view.

I think this conversation has been helpful -- is it okay if we expand it for a moment to consider some outside-the-box ideas such as:
- maybe not showing a row percentage? Maybe the line numbers are enough to give context?
- maybe using some in-window space for ascii scroll bar (-ish) UI elements (since they show the same context as percentage indicators).
- maybe showing no percentages at all
- maybe showing no percentages but instead showing only words or blank space (still show 'top' and 'bottom'; and maybe 'left' and 'right' as well)

aaxu:
> maybe not showing a row percentage? Maybe the line numbers are enough to give context?
I think the percentages give off information that just the line numbers don't give. So I don't think it's good to completely remove this "information"-giving feature.

> maybe using some in-window space for ascii scroll bar (-ish) UI elements (since they show the same context as percentage indicators).

Indeed if we are able to make both a horizontal and vertical scrollbar, then there would be no need for the percentages. I think we should end up creating this either way in the future just because it allows for quicker scrolling and provides an alternative to mouse scrolling.

> maybe showing no percentages at all
I was thinking perhaps we could simply show something like "currentLine / totalLines". This would make more sense and look less odd. For example, being on the first line of 4 lines would be 1/4.. then 2/4, 3/4, etc. Again, there is no 0/4 but I think it looks a bit better. Meanwhile for columns, it would go from 0/4 to 4/4. 

> maybe showing no percentages but instead showing only words or blank space (still show 'top' and 'bottom'; and maybe 'left' and 'right' as well)
I'm not really sure how this works. Do you mean completely remove percentages and only show top/bottom/left/right? I feel like this would give us very little information. It would tell us why we can no longer scroll in a direction, but it would also not really be relevant when we are in the middle of the file.

dschuyler:
>>maybe showing no percentages at all

>I was thinking perhaps we could simply show something like "currentLine / totalLines". This would make more sense and look less odd. For example, being on the first line of 4 lines would be 1/4.. then 2/4, 3/4, etc. Again, there is no 0/4 but I think it looks a bit better. Meanwhile for columns, it would go from 0/4 to 4/4.

Maybe just "totalLines" since the line numbers show the current line?

>> maybe showing no percentages but instead showing only words or blank space (still show 'top' and 'bottom'; and maybe 'left' and 'right' as well)

> I'm not really sure how this works. Do you mean completely remove percentages and only show top/bottom/left/right? I feel like this would give us very little information. It would tell us why we can no longer scroll in a direction, but it would also not really be relevant when we are in the middle of the file.

Yep, that's that idea. That's why it would be bank when not at one of those extremes. I was just brainstorming on why I look at those percentages, and it's really about those extremes. I'm not generally concerned if I'm 30% or 35% into the file.

dschuyler:
Can we try this one for awhile:

Think of row and column percentages as 'how many of the rows/columns are above/left'.

```
|    0%, 0%
``` 

```
x|    0%, 100%
``` 

```
|x    0%, 0%
``` 

```
|xxxx
xxxx
xxxx
xxxx    0%, 0%
``` 


```
x|xxx
xxxx
xxxx
xxxx    0%, 25%
``` 

```
xxxx
xx|xx
xxxx
xxxx    25%, 50%
``` 

```
xxxx
xxxx
xxx|x
xxxx    50%, 75%
``` 

```
xxxx
xxxx
xxxx
xxxx|    75%, 100%
```

dschuyler:
Trivia:
vim seems to use the percentages for what is shown on screen (not the cursor position).
So if all the lines are on screen it says 'all'.
If the first line is at the top of the screen it says 'top'.
When scrolled to the bottom (and there's stuff scrolled off the top) it says 'bottom'.
When not at the top or bottom it *might* show the percentage of lines not shown. I'm not sure. With say 39 lines on screen of a 49 line file, if the top line is 6, it says 50%.

aaxu:
> Maybe just "totalLines" since the line numbers show the current line?

I feel like it would be better to show both, since firstly it makes it more clear to the user what it actually is. For example, as they move their cursor around, they will see the currentLine change accordingly while totalLine will stay the same. So it would look like..
50/100
51/100
52/100.

We could do the same for columns as well. This would also make it so the user doesn't have to look between the status bar and the line numbers to determine what position they are at. Perhaps we could even do something like
53/100 (53%)

> Yep, that's that idea. That's why it would be bank when not at one of those extremes. I was just brainstorming on why I look at those percentages, and it's really about those extremes. I'm not generally concerned if I'm 30% or 35% into the file.

For this, I was thinking that it would just be nice to have an indicator of where we are at at the file. This could be useful for people who have a general sense of what position in the file they want to edit. Also, since I'm part of the developing (hence I do random things), I scroll to the end of the file and sometimes wonder how much longer till I get there, so that's probably the only time I've used this feature thus far.

aaxu:
I've changed the implementation to the one you wanted to try, but also let me know what you think of the ideas above. We can perhaps move the discussion details to a github issue.

app/window.py
```diff
@@ -557,15 +557,16 @@ def render(self):
...
-      charCount = len(tb.lines[self.host.cursorRow])
-      if (self.host.cursorCol < charCount):
-        colPercentage = self.host.cursorCol * 100 / charCount
-      else:
-        colPercentage = 100
+    # Assumes lineCount is always at least 1.
+    if lineCount == 1:
+      rowPercentage = 100
+    else:
+      rowPercentage = (self.host.cursorRow + 1) * 100 / lineCount
```
dschuyler: Oh, I think what I was asking about was not doing the +1, so row and column would both be the same equation: number * 100 / count. It means that column 100% is not reachable.

But I'll try this for awhile as-is. Then maybe we can try what I was wondering about.

dschuyler:
I'm going to land this to try it out for a bit. I like the idea of continuing this in a git bug.
After we/I have tried this for a bit, maybe we'll try another. It helps me to experience a change for a bit to see how it 'feels'.

PR #148, opened 2018-02-04
Refactoring/bookmarks

aaxu:
Just putting this out here for now so people can take a look. I know the other PR would probably also be changed if this gets pulled in first, or this will need to be changed if the other one gets pulled in first. I'll fix the changes if that happens. 

I created a new bookmark class which keeps track of a (begin, end) tuple which acts as an inclusive interval. All comparisons between bookmarks will only compare their intervals (since this would be useful when using `bisect`. Note that this means that two bookmarks with different data but equal ranges will be 'equal' to each other).

### DIFF
--- a/app/bookmark.py
+++ b/app/bookmark.py
@@ -76,3 +76,9 @@ class Bookmark:
 
   def __ge__(self, other):
     return self.range >= other.range
+
+  def __hash__(self):
+    return hash(self.range)
+
+  def __repr__(self):
+    return repr(self.range)
--- a/app/bookmark.py
+++ b/app/bookmark.py
@@ -24,8 +24,8 @@ class Bookmark(object):
       beginRow (int): The line that the bookmark starts on (inclusive).
       endRow (int): The line that the bookmark ends on (inclusive).
       data (other): This is used to store any information that you would like to
-                    to associate with this bookmark. It can be accessed by calling
-                    bookmark.getData()
+                    to associate with this bookmark. It is an empty dictionary
+                    by default.
     """
     self.__begin = beginRow
     self.__end = endRow
--- a/app/bookmark.py
+++ b/app/bookmark.py
@@ -1,3 +1,17 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 class Bookmark:
   """
   This bookmark object is used as a marker for different places in a
@@ -15,6 +29,14 @@ class Bookmark:
     """
     self.range = (beginRow, endRow)
     self.data = data
+
+  @property
+  def begin(self):
+    return self.range[0]
+
+  @property
+  def end(self):
+    return self.range[1]
 
   def overlap(self, bookmark):
     """
--- a/app/bookmark.py
+++ b/app/bookmark.py
@@ -55,29 +55,37 @@ class Bookmark:
     Returns:
       True if the passed in row is inside the bookmark's range.
     """
-    assert type(row) == int
+    assert isinstance(row, int)
     begin, end = self.range
     return begin <= row <= end
 
   def __lt__(self, other):
+    assert isinstance(other, Bookmark)
     return self.range < other.range
 
   def __gt__(self, other):
+    assert isinstance(other, Bookmark)
     return self.range > other.range
 
   def __eq__(self, other):
+    assert isinstance(other, Bookmark)
     return self.range == other.range
 
   def __ne__(self, other):
+    assert isinstance(other, Bookmark)
     return self.range != other.range
 
   def __le__(self, other):
+    assert isinstance(other, Bookmark)
     return self.range <= other.range
 
   def __ge__(self, other):
+    assert isinstance(other, Bookmark)
     return self.range >= other.range
 
   def __hash__(self):
+    # NOTE: Any two bookmarks with the same range WILL have the same hash value.
+    # self.range can also change, so be careful when using this in a hash table.
     return hash(self.range)
 
   def __repr__(self):
--- a/app/bookmark.py
+++ b/app/bookmark.py
@@ -23,8 +23,8 @@ class Bookmark(object):
     Creates a bookmark located at the specified rows. This will automatically
     cast the passed in rows into integers.
     Args:
-      beginRow (int): The line that the bookmark starts on (inclusive).
-      endRow (int): The line that the bookmark ends on (inclusive).
+      beginRow (int-like): The line that the bookmark starts on (inclusive).
+      endRow (int-like): The line that the bookmark ends on (inclusive).
       data (other): This is used to store any information that you would like to
                     to associate with this bookmark. It is an empty dictionary
                     by default.
@@ -39,7 +39,7 @@ class Bookmark(object):
 
   @range.setter
   def range(self, value):
-    self.__begin, self.__end = int(min(value)), int(max(value))
+    self.__begin, self.__end = sorted(int(x) for x in value)
 
   @property
   def begin(self):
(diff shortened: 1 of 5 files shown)


app/bookmark.py
```diff
@@ -0,0 +1,113 @@
...
+      row (int): the row that you want to check.
+
+    Returns:
+      True if the passed in row is inside the bookmark's range.
+    """
+    return self.begin <= row <= self.end
+
+  def __lt__(self, other):
+    assert isinstance(other, Bookmark)
+    return self.range < other.range
```
Androbin: This will compare the `begin` values first and `end` values second, right?
Would it make sense to include functions to compare only one or the other?
Or should this be rather done explicitly on the calling site?

Androbin:
How would I best rewrite https://github.com/google/ci_edit/pull/147/commits/c1231fd201b97d16247cd3edb36ed44326170225 on top of this?

app/bookmark.py
aaxu: Yes, by default all comparison operators will compare first the begin values then the end values. This should make it so that existing functions can simply be changed to use Bookmarks  since we were comparing using ranges (tuples) before too. If you want to compare only one value, you should do it explicitly by grabbing the values yourself. 

The only time I could think of where you would need to compare 9ne value explicitly is to check if either two bookmarks have the same begin or the same end. Otherwise, it would output the same answer as an 'implicit' (< > ==) check.

Androbin: Yeah, it's just that the other PR doesn't work with this behavior.
More specifically, if `begin < upper < end`, it's skipping the bookmark.

aaxu:
I think for your code, your `upperLimit` would just have to become a Bookmark with whatever value its supposed to bisect at. 

All references to some `bookmark[0]` should be changed to `bookmark.range`
bookmark[1] -> bookmark.data
bookmark[0][0] -> bookmark.begin 
bookmark[0][1] -> bookmark.end
range is also just a tuple so you can do
begin, end = bookmark.range

Androbin:
Yeah, it's just that the other PR doesn't work with this behavior.
More specifically, if `begin < upper < end`, it's skipping the bookmark.

aaxu:
Though you should wait until this is PR is accepted first before updating your PR.

aaxu:
I just tested the code on your branch and it seems to be working fine with this PR. I only did it quite quickly as I'm not completely sure of how the `bookmarkMove` function actually works, but it seemed to retain the same functionality. 

You can take a look at it and play around with it here: https://github.com/aaxu/ci_edit/tree/androbin/bookmark

Although I think I may have found a 'bug' with your branch, which I'll post a follow up on your PR.

Androbin:
Yeah, it works with the code from c1231fd201b97d16247cd3edb36ed44326170225.
It has the bug described above using `bisect.bisect_left(self.bookmarks, Bookmark(upper, upper))`.

aaxu:
Oh I see what you mean. In that case, we could do something like this maybe?
```
begin = bisect.bisect_left(self.bookmarks, Bookmark(upper, upper)) - 1
# Check if self.bookmarks[begin] 'shows up on screen', ie: self.bookmark[begin].end >= self.view.scrollRow.
# If so, move this bookmark, else skip.
```

Androbin:
Hmm, you mean because we won't encounter overlapping bookmarks?
If so, we'd still have to `begin = max(... - 1, 0)`, right?

aaxu:
Yes, dealing with intervals that don't overlap makes things much easier, so I think the implementation above would work, with the `begin = max(... - 1, 0)`.

Androbin:
Shouldn't the `Bookmark.data` keys be fields?

aaxu:
I'm not too sure what you mean. It's just a dictionary, which you can set to store any information you want. It's not necessarily any field intrinsic to the Bookmark object itself, if that answers your question.

Androbin:
Let me put it this way: Does this dictionary have a well known set of keys?

Note: I am specifically asking because of the "old positions" thing.
They should be about as "intrinsic" as `begin` and `end`.

aaxu:
I'm not sure. I needed the `begin` and `end` fields in order to implement the comparison functionality, and it was the minimum that I needed. I don't know if we want to hardcode the other fields into this bookmark object. We essentially do this all in the `dataToBookmark` function, so I don't see a reason to hardcode it in to the class itself. This would also let us give us one less point to maintain, in case we decide to add more data to the dictionary. What do you think?

Androbin:
Hmm, since the data in question is both declared and used/changed in `actions.py`, I think we are fine.
Otherwise it would have been nice to have that function inside the `Bookmark` class.

aaxu:
Branch should be up to date now. I fixed the bookmark tests to use the new bookmark object. Passing tests are still passing.

Androbin:
I see that `float('inf')` _could_ be useful. But I ain't quite sure how.
Would rather not have any part of the code have to deal with potential infinities.

aaxu:
That is for the case if we have a bookmark (x, y) and our needle begins at x, then the only way to guarantee that bisect_right goes past (x, y) is if our needle ends after y. This is just so that we don't have to check if there is a bookmark after where our needle ended and don't have to check if that bookmark should be included or not.

I don't think the infinity here poses any issues since it is never saved as a state and used only for the needle, but I'm willing to change it to the extra check implementation if you're still opposed to it.

Androbin:
We are _currently_ only doing comparison, no arithmetic.
As long as don't need `length = end - begin`, we should be fine.

aaxu:
Okay I've actually just removed the inf. It turns out there's no need for extra case checks.

aaxu:
Bookmarks will now automatically cast range values to integers (in case we do arithmetic and end up passing a floating point number in)

Androbin:
What is left to do for this PR?

aaxu:
Just waiting for it to undergo review

app/bookmark.py
```diff
@@ -0,0 +1,115 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+class Bookmark(object):
+  """
+  This bookmark object is used as a marker for different places in a
+  text document. Note that because text buffer lines index at 0, all
+  references to rows also assume that 0 is the first line.
```
dschuyler: Trivia: that's part of the reason I use the term 'row' rather than 'line number'. I use row (or index) for a zero based value; and 'line' for a one based value. i.e. (row = line - 1), which I hope is true throughout the code (please let me know if we're being inconsistent somewhere).

app/bookmark.py
```diff
@@ -0,0 +1,115 @@
...
+  """
+  def __init__(self, beginRow, endRow, data={}):
+    """
+    Creates a bookmark located at the specified rows. This will automatically
+    cast the passed in rows into integers.
+    Args:
+      beginRow (int): The line that the bookmark starts on (inclusive).
+      endRow (int): The line that the bookmark ends on (inclusive).
```
dschuyler: Is there a (type) specifier for 'thing that converts to int'? i.e. it's only kinda accurate that an int is expected -- sure an int is preferred, but it looks like anything that can convert to an int can be passed in. 
Or, maybe truly expect/require ints and not do the int() conversion below?
(Either way is okay with me, just making a note that the docs and code don't seem to line up).

Androbin: I think the python nomenclature would be 'int-like'.

app/bookmark.py
```diff
@@ -0,0 +1,115 @@
...
+    self.__end = int(endRow)
+    self.data = data
+
+  @property
+  def range(self):
+    return (self.begin, self.end)
+
+  @range.setter
+  def range(self, value):
+    self.__begin, self.__end = int(min(value)), int(max(value))
```
dschuyler: While I think it's unlikely to be an issue, is it possible that min/max() may pick a different value than they would if they were ints? e.g. (A, B) where A < B, but int(A) >= int(B).

WDYT of 
```
self.__begin, self.__end = int(value[0]), int(value[1])
if self.__begin > self.__end:
  self.__begin, self.__end = self.__end, self.__begin
```

Androbin: Why not just `sorted(value)` ?

Androbin: Or rather `sorted(int(x) for x in value)` ?

aaxu: > While I think it's unlikely to be an issue, is it possible that min/max() may pick a different value than they would if they were ints? e.g. (A, B) where A < B, but int(A) >= int(B).

Not that I know of? If A < B, and both are of the same sign, then they must both round toward the same direction when casted to an int. If they have differing signs, then int(A) will be <= int(B) since A <= 0 and B >= 0, they'll both round towards 0 (but never switch signs).

Unless there's some weird way of representing floating point numbers like in java that I don't know about..

dschuyler: Maybe something silly like
min('4', 5)
5 < '4', but int('4') < int(5)
It seems unlikely, but it also seems possible. Or said another way, with enough time and use of some code, someone is likely to eventually trip over that situation.

aaxu: ooo I see. Okay I just fixed the setter functions. What do you think about it now?

Androbin: In such cases, Python would give:
`TypeError: unorderable types: int() < str()`

PR #151, opened 2018-02-10
dictionary words

aaxu:


app/dictionary.en-us.words
```diff
@@ -10996,6 +11008,7 @@ nowadays
 noway
 nowhere
 nowt
+ntfs
```
dschuyler: Thanks!, Hey please move this to 
ci_edit/app/dictionary.acronyms.words

P.S. I've been wondering about suggesting that acronyms be capitalized (since that would be proper). To keep that option open, I'm trying to keep the acronyms separate. Trivia, I'm also aware that many of them are actually initialisms and not acronyms; sorry if that's confusing.

app/dictionary.en-us.words
```diff
@@ -10996,6 +11008,7 @@ nowadays
...
 nth
 ntp
```
dschuyler: Optional (since this wasn't added in this PR, but hey I just noticed that ntp should be in ci_edit/app/dictionary.acronyms.words as well, if you're up for moving it).

app/dictionary.en-us.words
aaxu: Sure thing!

app/dictionary.en-us.words
aaxu: Added

PR #156, opened 2018-04-11
Tests for automatic column adjustment feature

aaxu:


dschuyler:
Thanks for creating this.

Tip: it's nicer to get smaller, focused PRs. Like, 
- the dictionary changes could be a PR on their own
- moving the function about the test file could have been another PR
- adding the new tests could have been a third PR.

That said, these are all small; and I'm taking them anyway :)

bonus: Doing smaller, focused CLs will also help you use `git bisect` (which is very helpful).

dschuyler:
Heh, CL is for Change Request. PR is for Pull Request. They mean essentially the same thing. Sorry that I sometimes use one or the other. "pull" is a git/github thing. Like the generic category of thing that "git" is, is a Revision Control System. The generic name for a pull request is a change request. In my normal work, the term CL is used. When on github, I try to use PR, but sometimes I forget and write CL.

The "prefer small changes" is also encouraged at my work. I didn't make particularly small changes in the past, but now I do and I'm seeing the wisdom in it. For some (the initial paper I read on the topic) the advantage is in reviewing the change - keeping it small is more efficient for someone reading/approving the change. 

To me, the big win is bisecting the code to see where something broke for the purpose of fixing it. If git bisect narrows it down to a small change, it's much easier to work out the bug and fix it. Conversely, if git bisects down to a pile of changes that all happened together, it's less clear what caused the bug being investigated.

In places I've worked that were concerned with having too many small changes, they were optimizing for the log of the change history. So if that's the priority you'd want to do the opposite (e.g. combine changes together). I don't optimize for reading the history, but it's just a choice. 

Hmm, I think I should write up this choice so that others will know. I should make a style guide too.

aaxu:
Oh that's good to know. Having a style guide and contribution guide would be pretty nice as well! But... where is the L in "Change request"?

dschuyler:
Oh my, I wrote that too early in the morning. CL is Change List.

For the commits being grouped together. I've been using the "Squash and Merge" button (which is the default) to bring in PRs. It looks like there's also an option to bring in the changes as separate changes (without squashing them).

Overall, this may just be a thread of me mistyping things and pressing the wrong buttons to pull changes. Thank you for bearing with me.

aaxu:
Oh I see. :+1: Thanks!

PR #159, opened 2018-04-11
Added a length check in searchForward to prevent indexOutOfBounds

### LINKED ISSUE #157: Program crashes when you try to move cursor past a lone left parenthesis
Steps to reproduce:
* type ( into any file
* press the left arrow key

aaxu:
Resolves #157


aaxu:
oh haha we both fixed it at the same time.

PR #169, opened 2018-04-21
Fixed test runner

aaxu:
The commit "1d628d76" seemed to have changed the `runTests` function, but didn't change how calls to it were made. This should fix it. 

One question though: do you know why when I just run `python unit_tests.py`, it doesn't run the performance test, but when I put a run the same thing with a `import pdb; pdb.set_trace` break statement, and simply hit continue, it runs the performance tests?

### DIFF
--- a/unit_tests.py
+++ b/unit_tests.py
@@ -63,7 +63,7 @@ tests = {
 
 def runTests(tests, stopOnFailure=False):
   """Run through the list of tests."""
-  for test in tests:
+  for test in tests.values():
     suite = unittest.TestLoader().loadTestsFromTestCase(test)
     result = unittest.TextTestRunner(verbosity = 2).run(suite)
     if stopOnFailure and (result.failures or result.errors):
@@ -72,4 +72,4 @@ def runTests(tests, stopOnFailure=False):
 
 if __name__ == '__main__':
   app.log.info("starting unit tests")
-  app.log.wrapper(runTests)
+  app.log.wrapper(lambda: runTests(tests))


dschuyler:
Thanks for pointing this out! I normally run `we --test`. I need to run `python unit_tests.py` when I'm making changes in that area. 

The PR here fixed `python unit_tests.py`, but broke `we --test`. Rather than giving feedback on this PR I went and rearranged how the unit tests are run more broadly. (I didn't want to dump a whole bunch of changes on this PR).

So now using `we --test` or `python unit_tests.py` behave much more similarly.

With either, you may now pass `--log` to get app.log output (or not, by omitting --log).

Also, a particular set of tests can be run by passing the name of the test as the last argument. E.g. `python unit_tests.py parser` (or `we --test parser`). Not specifying a named set of tests will run them all.

dschuyler:
> do you know why when I just run python unit_tests.py, it doesn't run the performance test...

My guess would be that some other test was causing the unit_tests.py to exit before running them. I had some sys.exit calls in unit_tests.py that I just removed. Maybe that will change the behavior you're seeing.
If that's not it, I'm not sure.
I did just try running ./unit_tests.py and the performance tests were run for me.

aaxu:
Ohh I see.

Here is the weird thing that I can't figure out.
If you go into `unit_tests.py` and insert `import pdb; pdb.set_trace()` in line 105, once you run `python unit_tests.py`, you will enter the debugger in your terminal.

At this point, if you enter 'c', you will finish the tests to completion at a significantly faster rate than if you enter 'n'. `test_splice_vs_startswith` fails when you use 'n' rather than 'c'. 

It was my mistake about some performance tests being run. They all do run. It was just the issue above.

dschuyler:
> At this point, if you enter 'c', you will finish the tests to completion at a significantly faster rate than if you enter 'n'.

My first answer is I don't know. I don't recall noticing this before.

The docs at https://docs.python.org/2/library/pdb.html says that `next` will continue at "(nearly) full speed", which indirectly implies that isn't not really full speed. The `continue` command simply continues until a break is hit (with no mention of a speed impact).

I like to hypothesize, so here's a guess: it's common for source debuggers to install interrupt handlers to help control/monitor the program being controlled by the debugger. The ci_edit application also makes use if interrupts. Maybe there's extra overhead involved in both things handling interrupts. Again the proceeding is a guess, I didn't dig into it. To dig into it (if you're so inclined) you could read the pdb source (time consuming), ask on a Python forum (potentially long wait and uncertain reliability of the answer), or set up empirical tests (may take a few hours to isolate what causes the slow down). But in the end there's a potential difference in performance that's unlikely to change :), sorry.

aaxu:
Haha I'll just accept this for now. No problem. Thanks for fixing this.

PR #181, opened 2018-12-25
Implement jump to opposite bracket feature

sauravchirania:
Do we have a better choice than CTRL_B key?
Is there a rule to arrange methods inside a class? I have created two new methods just below __init__() in Actions class. Should I move them somewhere lower in the file?

### DIFF
--- a/app/actions.py
+++ b/app/actions.py
@@ -62,9 +62,9 @@ class Actions(app.mutator.Mutator):
         self.debugUpperChangedRow = -1
         self.parser = app.parser.Parser()
         self.fileFilter(u'')
-    
-    def getMatchingBracketAddress(self):
-        """Returns the address of the bracket which matches
+
+    def getMatchingBracketRowCol(self):
+        """Gives the position of the bracket which matches
         the bracket at the current position of the cursor.
 
         Args:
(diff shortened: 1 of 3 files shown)


dschuyler:
Sorry for the delay. I was out on vacation.

We're pretty limited on key combinations in the terminal. I'd prefer something like ctrl+m for matching. Unfortunately ctrl+m can be tricky because it may be mapped to ctrl+j.

Order of methods: I generally try to keep the methods in alphabetical order. Though I haven't always stuck to this, so it's a weak rule.

app/actions.py
```diff
@@ -62,6 +62,76 @@ def __init__(self, program):
...
+
+        Args:
+          None.
+
+        Returns:
+          None if matching bracket isn't found.
+          Address (int row, int col) of the matching bracket otherwise.
+        """
+        if (self.parser.rowCount() > self.penRow and
+                len(self.parser.rowText(self.penRow)) > self.penCol):
```
dschuyler: Unless there's a reason not to: 
Try to reverse the logic in lines 77 and 78, then do `return None` as the body of the `if`; and unindent the rest of the function.
(There are two competing schools of thought here: one is to minimize the number of `return` statements and the other is to early out and keep the normal flow more like a list (keep code to the left column) than nesting code (indenting to the right). I believe the latter is a better choice.)

app/actions.py
```diff
@@ -62,6 +62,76 @@ def __init__(self, program):
         self.debugUpperChangedRow = -1
         self.parser = app.parser.Parser()
         self.fileFilter(u'')
+    
+    def getMatchingBracketAddress(self):
```
dschuyler: What do you think of changing 'Address' to 'RowCol'? Address is okay if you prefer to keep 'Address', but it takes me and extra moment to realized that it doesn't mean a memory address (e.g. 0x0000000043238080) where 'RowCol' I don't have to take that moment.

sauravchirania: RowCol looks better to me too.

app/actions.py
sauravchirania: Fixed

sauravchirania:
Thank you for the code review :-)

> We're pretty limited on key combinations in the terminal. I'd prefer something like ctrl+m for matching. Unfortunately ctrl+m can be tricky because it may be mapped to ctrl+j.

Yes, ctrl+m looks like a better idea. I'll see if I can figure out a way to map this feature to ctrl+m.

dschuyler:
We can also land this CL without specifying ctrl+b for matching. E.g. comment out that CTRL_B line in cu_editor.py.

PR #187, opened 2019-01-18
Make status bar sensitive to selection mode changes

### LINKED ISSUE #186: Delay in updation of status bar
This is to continue issue #182. Since I couldn't reopen it, I am starting a new issue.

Some work has already been done in 72fd6d4 to fix it, but I believe that the status bar shouldn't only get cleared on pressing backspace or carriage return, but also on pressing of any other key that deletes the selected text.

Steps to reproduce:

Open ci_edit and type some text.
Use CTRL+A to select all text (status bar also gets updated to show the current selection mode)
Press the 'a' key.

This deletes all the selected text and we are left with only the character 'a' written on screen. But, status bar still shows the previous selection mode (select all). I would expect the status bar to show no selection. @dschuyler , what do you think?

sauravchirania:
This is to fix #186.


dschuyler:
Good change. I'm up for merging it as-is, but I wanted to ask if you're up for trying your hand at writing a test (to see that this doesn't get broken again in the future). 

Maybe something like test_save_on_close() in app/unit_test_ui.py. (Easiest and recommended -- this is actually a higher level test (not a unit test)).
Or a new test class in app/unit_test_actions.py (Harder, there would be hookup needed in unit_tests.py as well and doesn't test what is displayed on screen).

If you're not up for writing a test that's ok (but a test would be great).

sauravchirania:
Thanks!
I am interested in writing tests too. I'll send a patch soon.

sauravchirania:
Hi!
I am sorry for the delay. It was because I had to participate in two contests last weekend, and also because it took me time to understand how the testing is working.

> Maybe something like test_save_on_close() in app/unit_test_ui.py.

When I run this test using `./unit_tests.py ui`, I get the following error -

======================================================================
FAIL: test_save_on_close (app.unit_test_ui.UiBasicsTestCases)

Traceback (most recent call last):
  File "/home/intelcore/ci_edit/app/unit_test_ui.py", line 128, in test_save_on_close
    CTRL_Q
  File "/home/intelcore/ci_edit/app/fake_curses_testing.py", line 281, in runWithTestFile
    self.runWithFakeInputs(fakeInputs, ["ci_test_program", kTestFile])
  File "/home/intelcore/ci_edit/app/fake_curses_testing.py", line 259, in runWithFakeInputs
    self.prg.run()
  File "/home/intelcore/ci_edit/app/ci_program.py", line 440, in run
    self.commandLoop()
  File "/home/intelcore/ci_edit/app/ci_program.py", line 187, in commandLoop
    ch = cursesWindow.getch()
  File "./test_fake/curses/curses.py", line 300, in getch
    val = fakeInput.next()
  File "./test_fake/curses/curses.py", line 82, in next
    result = cmd(self.fakeDisplay, self.inputsIndex)
  File "/home/intelcore/ci_edit/app/fake_curses_testing.py", line 239, in copyToClipboard
    self.assertTrue(clipboard.copy, callerText)
AssertionError: <third_party.pyperclip.clipboards.ClipboardUnavailable object at 0x7f18d23a4850> is not true : in unit_test_ui.py:122:test_save_on_close(): 

----------------------------------------------------------------------

dschuyler:
That sounds like determine_clipboard() is failing. See ci_edit/third_party/pyperclip/__init__.py around line 49. The error suggests that it didn't find any clipboard (for copy/paste) match for your OS and returned `return init_no_clipboard()` on line 83.

So copy/paste should still work _within_ ci_edit, but it might not work between other windows and ci_edit. If you try to copy/paste between ci_edit and some other program (like your browser) does it work?

sauravchirania:
Copy/paste within ci_edit works as expected (ctrl+c for copy, ctrl+v for paste). If I copy something from other program and press ctrl+v to paste it in ci_edit, it doesn't get pasted. But if I use ctrl+shift+v to paste it, it works fine.

sauravchirania:
I found the solution here - https://pyperclip.readthedocs.io/en/latest/introduction.html#not-implemented-error

After I installed the xsel utility, the tests as well as the copy-paste is working as expected.

dschuyler:
Thanks!

PR #189, opened 2019-01-31
Feature/ctrl backspace

aaxu:
Finally got around to figuring this out. I've only tested this out on Ubuntu 18.04 since I currently don't have access to a Mac, though I will by next week. More testing would be appreciated but here's a quick look at what I have so far. 

Main points:
* The program should be able to handle ctrl_backspace differently from other backspaces (ie Backspace, and CTRL_H)
* This feature should not work on macOS because of security issues, and it should simply behave as a normal backspace.
* This feature should not work if there is no X server running, so it may crash when accessed over an ssh server (needs testing)


aaxu:
Update: Program should no longer crash when there is no X server available. It will act as if this feature did not exist and CTRL_Backspace should just be a normal backspace.

app/actions.py
```diff
@@ -1094,11 +1113,10 @@ def scrollToOptimalScrollPosition(self):
...
             optimalRowRatio = self.program.prefs.editor[u'optimalCursorRow']
-            scrollRow = max(
-                0,
-                min(
-                    len(self.lines) - 1,
-                    top - int(optimalRowRatio * (maxRows - 1))))
+            scrollRow = max(0,
+                            min(
+                                len(self.lines) - 1,
+                                top - int(optimalRowRatio * (maxRows - 1))))
```
dschuyler: I'm guessing some of these diffs are from different style settings in yapf. Do these change when using `yapf -i --style google <file>`.
Also, there's a tool at `ci_edit/tools/presub` that should do the yapf on changed files and run pylint that might be helpful.

app/ci_program.py
```diff
@@ -442,6 +445,14 @@ def run(self):
             self.bg.put((self.programWindow, 'quit'))
             self.bg.join()
 
+    def setUpKeyboardMonitor(self):
+        try:
+            self.keyboard_monitor = app.keyboard_monitor.KeyboardMonitor()
+            self.keyboard_monitor.start()
+        except:
+            # In case of a crash, do nothing.
```
nthparameter: Consider saying "exception" or "failure" rather than crash.

app/keyboard_monitor.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class KeyboardMonitor:
+    """
+  This class is used to listen to the user's inputs in order to detect any
+  modifier keys being used. Due to limitations of this library, errors may occur
+  when running over ssh, since an X server is required. This will also not be
+  able to detect most keys on macOS unless run as root. See more details at
+  https://pynput.readthedocs.io/en/latest/limitations.html.
+  """
```
nthparameter: yapf doesn't indent the doc string. So this should be indented a bit.

app/actions.py
nthparameter: Hmm, the style in this CL is actually nicer than the --style google settings. Maybe we should consider not using --style google.

app/controller.py
```diff
@@ -19,6 +19,10 @@
```
nthparameter: See note in app/keyboard_monitor.py

app/keyboard_monitor.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2019 Google Inc.
```
nthparameter: Let's land this PR with out app/controller.py changes or app/keyboard_monitor.py. The pynput library looks like it's GPL, that can be problematic in some uses. If we do two PRs we can land the other changes sooner.

aaxu: Hmm, yeah I see the conflicts with the licenses. I'll have to see if it's possible to do the detection myself or find some alternative library..

dschuyler: I'm happy to land the rest of this CL if you're up for clipping the app/controller.py and app/keyboard_monitor.py changes (the other changes have no license issues).

aaxu:
I've just removed the parts using pynput, keeping only the functions that support backspacing a word. Currently, there will be no way to actually activate this function since it isn't mapped to any key combination.

PR #190, opened 2019-04-11
RFC: Added 'open' command to open a file under the cursor.

gennad:
Hello, this is my first commit to the project.

As it is my first commit, I'm pretty rough in terms of best practices so any tips, hints and comments will be greatly appreciated.

As was advised in email dev list, I implemented the command to open a file under cursor.

I tested it on a filename that does not exist (it output an error message) and tested on the filename that does exist (it opened but discarded the existing buffer)

### DIFF
--- a/app/interactive_prompt.py
+++ b/app/interactive_prompt.py
@@ -107,6 +107,7 @@ class InteractivePrompt(app.controller.Controller):
             u'cua': self.changeToCuaMode,
             u'emacs': self.changeToEmacsMode,
             u'make': self.makeCommand,
+            u'open': self.openCommand,
             #u'split': self.splitCommand,  # Experimental wip.
             u'vim': self.changeToVimNormalMode,
         }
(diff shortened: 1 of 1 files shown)


app/interactive_prompt.py
```diff
@@ -171,6 +172,38 @@ def noOp(data):
...
 
+    def openCommand(self, cmdLine, view):
+        """
+        Opens the file under cursor.
+        """
+        penLine = view.textBuffer.lines[view.textBuffer.penRow]
+        wordStart = wordEnd = view.textBuffer.penCol
+
+        # Locate the boundaries of the word under the cursor
+        while wordStart > 0 and penLine[wordStart - 1] != ' ':
```
dschuyler: How about looking for ", ', :, and <, > instead of space.
Consider the use case of .cpp include:
#include "foo/bar/blah.h"
#include <thing/stuff.h>

app/interactive_prompt.py
```diff
@@ -171,6 +172,38 @@ def noOp(data):
...
+
+        penLineLength = len(penLine)
+
+        while wordEnd < penLineLength and penLine[wordEnd] != ' ':
+            wordEnd += 1
+
+        selectedWord = penLine[wordStart: wordEnd]
+        expandedPath = os.path.abspath(os.path.expanduser(selectedWord))
+
+        if not os.path.exists(expandedPath) or not os.path.isfile(expandedPath):
```
dschuyler: Will this cover the current cases and check read access?
if not os.access(expandedPath, os.W_OK) or not os.path.isfile ...
The case I'm concerned about is a file that exists, but we don't have read permission for (chmod -w foo).

app/interactive_prompt.py
gennad: Maybe `isalnum`?
https://docs.python.org/3/library/stdtypes.html#str.isalnum

app/interactive_prompt.py
gennad: Agree, will modify

app/interactive_prompt.py
dschuyler: That will reject a period/dot so 'hello.h' would be an issue.
```
>>> u"hello.h".isalnum()
False
```
I'm guessing this will need to be done by hand and require ongoing tweaking; or integration with the parser. E.g. maybe the parser could be queried for the path under the cursor. 

That would be an advanced change though. So I wouldn't recommend digging in there unless your really interested in getting deep into it. The code for the parser is a bit weird for performance reasons. 

I'm going to suggest we commit this PR without those tweaks or parser changes. The PR is an advancement over what we had before. Those tweaks or parser changes can happen in future PRs.

(Trivia: I'd intended to prototype the editor in Python and migrate to a systems language when Python became unsuitable (not fast enough), but I keep being able to make Python work so well that the need to change hasn't been needed; though that means the parser is a touch unusual).

app/interactive_prompt.py
```diff
@@ -171,6 +172,44 @@ def noOp(data):
...
+        penLine = view.textBuffer.lines[view.textBuffer.penRow]
+        wordStart = wordEnd = view.textBuffer.penCol
+
+        # Locate the boundaries of the word under the cursor
+        while wordStart > 0 and penLine[wordStart - 1] not in fileSeparators:
+            wordStart -= 1
+
+        penLineLength = len(penLine)
+
+        while wordEnd < penLineLength and penLine[wordEnd] != ' ':
```
dschuyler: Did ya mean to change `penLine[wordEnd] != ' '` to
`penLine[wordEnd] not in fileSeparators`
so that it works similarly to line 189?

PR #193, opened 2019-04-13
Implemented format filter for Python

gennad:
Implemented format filter for Python using black formatter


app/formatter.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2016 Google Inc.
```
dschuyler: 2019

app/interactive_prompt.py
```diff
@@ -152,20 +153,29 @@ def focus(self):
...
-        lines = self.view.host.textBuffer.doDataToLines(
-            formatter.get(ext,
-                          noOp)(self.view.host.textBuffer.doLinesToData(lines)))
+
+        formatter = formatters.get(ext)
+
+        if not formatter:
+            return lines, u'No formatter for extension {}'.format(ext)
+
+        text = self.view.host.textBuffer.doLinesToData(lines)
```
dschuyler: Let's use ...textBuffer.linesToData(lines). I acknowledge that right now it's the same (it just calls through to doLinesToData).
Someday I plan to support editing binary files and at that point it may matter.

app/interactive_prompt.py
```diff
@@ -152,20 +153,29 @@ def focus(self):
...
+
+        try:
+            formattedText = formatter(text)
+        except RuntimeError as err:
+            return lines, str(err)
+
+        lines = self.view.host.textBuffer.doDataToLines(formattedText)
```
dschuyler: textBuffer.dataToLines

app/interactive_prompt.py
```diff
@@ -152,20 +153,29 @@ def focus(self):
         self.textBuffer.selectionAll()
 
     def formatCommand(self, cmdLine, lines):
-        formatter = {
+        formatters = {
             #".js": app.format_javascript.format
-            #".py": app.format_python.format
             #".html": app.format_html.format,
+            ".py": app.formatter.formatPython
```
dschuyler: Just musing (no change needed): hmm, if we have a 'formatter' file (rather than 'format_python' etcetera) maybe this table should move to that file. WDYT (what do you think).
I'm also curious about (though fine with) the decision to go with a 'formatter' file rather than 'format_python'.
(Again, this is all okay for this PR, I'm just interested in your opinion).

PR #207, opened 2019-09-19
add pip package

### LINKED ISSUE #198: Add a package to pypi
To enable the installation via `pip install ci_edit`

fsx950223:
Fix #198

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,26 @@
+from setuptools import setup
+from setuptools import find_packages
+from datetime import datetime
+import io
+import os
+
+here = os.path.abspath(os.path.dirname(__file__))
+
+def get_long_description():
+  # Read the long-description from a file.
+  with io.open(os.path.join(here, 'readme.md'), encoding='utf-8') as f:
+    return '\n' + f.read()
+
+setup(
+    name='ci_edit',
+    version=datetime.strftime(datetime.today(), "%Y%m%d"),
+    description='A terminal text editor with mouse support and ctrl+Q to quit.',
+    long_description=get_long_description(),
+    long_description_content_type='text/markdown',
+    author='Google Inc.',
+    author_email='opensource@google.com',
+    url='https://github.com/google/ci_edit',
+    packages=find_packages(),
+    scripts=['ci.py'],
+    license='Apache 2.0',
+)
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # TODO(dschuyler): !/usr/bin/python -O
 
-# Copyright 2016 Google Inc.
+# Copyright 2019 Google Inc.
 #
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+# TODO(dschuyler): !/usr/bin/python -O
+
+# Copyright 2016 Google Inc.
```
dschuyler: Please update to 2019

setup.py
```diff
@@ -0,0 +1,47 @@
...
+  with io.open(os.path.join(here, 'readme.md'), encoding='utf-8') as f:
+    return '\n' + f.read()
+
+setup(
+    name='ci_edit',
+    version=datetime.strftime(datetime.today(), "%Y%m%d"),
+    description='A terminal text editor with mouse support and ctrl+Q to quit.',
+    long_description=get_long_description(),
+    long_description_content_type='text/markdown',
+    author='Google Inc.',
```
dschuyler: Hmm, so I'm ('Dave Schuyler') the original author. Though Google holds the copyrights it's not a 'google product'. If I'm the 'author' then the email I use for ci_edit is dschuyler@chromium.org

If ya feel differently, please let me know.

I tried looking at some other setup.py files in github.com/google.
These don't have author info, but has maintainers info.
https://github.com/google/timesketch/blob/master/setup.py
https://github.com/google/turbinia/blob/master/setup.py
I'm not sure if that's due to not having a single original author(?) or maybe maintainers imply that the author is no longer involved and is now being maintained by someone else?

This one has a single person for the author and maintainer
https://github.com/google/python-adb/blob/master/setup.py

setup.py
```diff
@@ -0,0 +1,47 @@
...
+    author_email='opensource@google.com',
+    url='https://github.com/google/ci_edit',
+    packages=find_packages(),
+    scripts=['ci.py'],
+    license='Apache 2.0',
+)
```
dschuyler: Looking at 
https://packaging.python.org/tutorials/packaging-projects/
I wondered about some classifiers, maybe:

    classifiers=[
        "Programming Language :: Python :: 3",
        "Environment :: Console",
        "Environment :: Console :: Curses",
        "License :: OSI Approved :: Apache Software License",
        "Topic :: Text Editors",
    ],

What do you think?

setup.py
```diff
@@ -0,0 +1,47 @@
...
+    return '\n' + f.read()
+
+setup(
+    name='ci_edit',
+    version=datetime.strftime(datetime.today(), "%Y%m%d"),
+    description='A terminal text editor with mouse support and ctrl+Q to quit.',
+    long_description=get_long_description(),
+    long_description_content_type='text/markdown',
+    author='Google Inc.',
+    author_email='opensource@google.com',
```
dschuyler: Maybe
dschuyler@chromium.org

PR #210, opened 2019-12-10
Added a way to run ci_edit directly on an online compiler, Repl.it

nraikm:
I have added a one click button, which will automatically set up ci_edit and run the editor in the Repl.it online environment.
image

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,2 @@
+language = "python"
+run = "python ci.py"
--- a/readme.md
+++ b/readme.md
@@ -1,3 +1,4 @@
+[![Run on Repl.it](https://repl.it/badge/github/google/ci_edit)](https://repl.it/github/google/ci_edit)
 # What is ci_edit
 
 ci_edit is a text editor. It can help you view or edit text files.


dschuyler:
Thanks. Hey, when I looked at ci_edit on replit, the cursor doesn't show up (it looks like it's white on a white background). Does it look that way for you?

Mosrod:
Yeah, it looks the same for me, and I believe it is due to the cursor being on a white background (as you have explained above).

dschuyler:
I like the idea of this PR. It gives potential users a chance to see (and use) ci_edit. With the white cursor, I'm concerned that someone would pass on using ci_edit. That would be sad.

Do you have time to look into the issue and see if we can get the cursor to show up (before we land this PR)?

Mosrod:
Repl.it uses a white cursor on the terminal, and I couldn't figure out a way to change it. I believe that changing the cursor would require changes to their environment, but there might be another way to do it. Perhaps in the future, there might be more console themes, which would change the cursor to the desired color. I agree that some people might pass on using ci_edit due to the white cursor, and maybe I could add a notice that would tell people that the cursor is white.

